PR #76, opened 2015-01-04
Support "touch" command for memcached and allow negative

mckelvin:
The [touch command](https://github.com/memcached/memcached/blob/master/doc/protocol.txt#L318) is neither supported in dynomite nor in twemproxy(cc @manju @idning ), I'm not sure if there's any deliberate purpose, if so, please tell me.

Touch with negative exptime is supported by memcached while not well documented. I'm not familiar with its inner implements but I guess the proxy should better transport without any interruption.

```
 telnet localhost 11211
Trying 127.0.0.1...
Connected to localhost.
Escape character is '^]'.
set foo 0 0 2
10
STORED
get foo
VALUE foo 0 2
10
END
touch foo -1
TOUCHED
get foo
END
```

### DIFF
--- a/src/dyn_message.h
+++ b/src/dyn_message.h
@@ -55,6 +55,7 @@ typedef enum msg_type {
     MSG_REQ_MC_PREPEND,
     MSG_REQ_MC_INCR,                      /* memcache arithmetic request */
     MSG_REQ_MC_DECR,
+    MSG_REQ_MC_TOUCH,                     /* memcache touch request */
     MSG_REQ_MC_QUIT,                      /* memcache quit request */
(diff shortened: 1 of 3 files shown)


blakesmith:
Any feedback on this @timiblossom? This would need to be merged with latest master, but this looks pretty straightforward to me.

timiblossom:
@mckelvin  and @blakesmith, we are running out of man power resources to support the features that we don't need currently.  If you really need this, please feel free to submit a patch.  Thanks.

timiblossom:
Sorry guys, this is already a patch.  I will review it soon.

mckelvin:
:beer:

Issue #85, opened 2015-02-06
Plugin Architecture for New Protocols


blakesmith:
Hey there!

I've been interested in Dynomite and would like to work on supporting a new protocol with it. I've been reading through the Dynomite source code, and the code base seems to be coupled strongly to redis and memcache (code paths such as: https://github.com/Netflix/dynomite/blob/master/src/dyn_dnode_msg.c#L446 and others).

Given dynomite's lineage with memcache, with redis added later, I would like to send patches that decouple the codebase from redis / memcache explicitly and roughly follow this development path:

1. Begin by decoupling the config from `redis` or not, and support `proto` as an arbitrary string in the config file. We can preserve backwards compatibility with the `redis` boolean if desired, but it'd be nice to deprecate it eventually if that's important to everyone.
2. Decouple internal `msg` branches that explicitly check for the `redis` boolean and other places that explicitly branch on memcache / redis. 
3. Put in place an interface for a more flexible "Protocol Interface" and move redis and memcache to this interface.
4. Move to allow dynamic linking to "Protocol Interfaces" that can be built outside of dynomite and decoupled from its build.
5. Investigate seeds providers for bootstrapping on these protocol interfaces? This is an area I'm fuzzy on, but could be useful to have built-ins for Protocol Interfaces that ship with Dynomite built-in.

Is there anything else I'm missing? Does this sound like a worthwhile effort? Having Dynomite be easily pluggable to other protocols would be the end goal. Let me know what you think!

Blake

timiblossom:
Hi Blake,

It is good to hear that you want to send out a patch.  Yanis already started to work on MongoDb and you can watch his work here at https://github.com/ipapapa/dynomite.  This should answer your questions #1, #2 and #3.

For #4, i think it probably needs that if some protocol requires a heavy dependency on other libraries.  Currently it is not an issue but will probably be in the future.

For #5, seed provider is just a place to return a list of seed nodes (or all of the nodes if we don't need gossip here).  Our current seed provider is dyn_seeds_provider is basically a mechanism to call a local REST interface.  You can base/extend on that to pull out seeds from other sources such as Mysql, a S3 file or a data on a shared storage.

blakesmith:
[...]

PR #86, opened 2015-02-07
Encapsulate request forwarding inside the protocol definition itself for messages that need replication

blakesmith:
Hey!

In a very first step towards the goal outlined in https://github.com/Netflix/dynomite/issues/85, this PR removes the `request_send_to_all_racks(struct msg *msg)` call inside `dyn_request` in favor of encapsulated protocol level functions that check if downstream replication is needed.

Also: can you think of a better name than `broadcast_racks`?

Let me know what you think,

Blake


blakesmith:
Closing this, since https://github.com/Netflix/dynomite/pull/87 changes the contract to `msg->is_read`.

PR #87, opened 2015-02-09
Add metrics for latencies

timiblossom:


src/dyn_histogram.c
```diff
@@ -0,0 +1,259 @@
...
+/* a port from this java code:
+ * https://github.com/apache/cassandra/blob/cassandra-1.2/src/java/org/apache/cassandra/utils/EstimatedHistogram.java
+ *
+ * Will try to use https://github.com/HdrHistogram/HdrHistogram_c  later
+ */
+
+
+
+static uint64_t bucket_offsets[BUCKET_SIZE];
+static uint64_t buckets[BUCKET_SIZE];
```
blakesmith: Will this be the only histogram in the codebase? Would it be better to have `buckets` on a struct for allocation in other parts of the codebase?

timiblossom: Hi Blake,

You are absolutely correct that we should make it more generic for other historgrams.  However, we need a quick measurement for coordinator latencies now.  Moving forward, we would rather use HdrHistogram as it is magically faster.  Even Cassandra is using HdrHistogram (java version) in its later versions

blakesmith: Great, thanks. :-)

Issue #42, opened 2014-11-05
Support for LEX and HLL commands in Redis


mattrobenolt:
I contributed these a while back to twemproxy and can easily be merged in here:

HLL: https://github.com/twitter/twemproxy/commit/0d3084be2c4d3d8cb64cb6199db3413d2cbcda42
LEX: https://github.com/twitter/twemproxy/commit/f0f6e6b08ca8c356f653b1998f0585aeb85779cb

Let me know if you'd like a specific Pull Request for these or if you can just yank these commits. Looks like they'd just apply cleanly.

ipapapa:
Did the work due to inactivity: https://github.com/Netflix/dynomite/pull/294/commits/4072346c90e4a1a6823bcbe40ffc0dc95d5d0748

PR #118, opened 2015-04-21
Added support for HyperLogLog's PFADD and PFCOUNT

mansoor-s:
Added HyperLogLog PFADD & PFCOUNT operations. 

Ported from the Twemproxy PR mentioned in this issue
https://github.com/Netflix/dynomite/issues/42

Note: PFMERGE wasn't added due to the fact that we can't guarantee that they keys live on the same node.

Note 2: PFCOUNT supports multiple keys. If multiple keys are specified Redis will return the union of their cardinality. Internally, it performs a PFMERGE operation. This is not supported here since we can't guarantee that the keys live on the same node.


ipapapa:
@mansoor-s can you please fix the conflicts and also make the commit towards the dev branch? We do not commit to master anymore.

ipapapa:
I am going to close this pull request, as the change has to potentially come to a new pull request in order to go to Netflix:dev.

PR #121, opened 2015-04-23
Updated docs to add clarification about compile dependencies, ports and ...

redboy1972:
...added a diagram from the forums.

NOTE: This is my first attempt to do something like this.  I know the image is remote but I don't know how to add it to the server.  Also am new to markdown.

Accept suggestions.  If this works out as I learn more about the product I will update pages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,8 @@ When Dynomite team decided to settle on using C language, we found Twemproxy and
 
 
 ## Build
+You may need to install the following dependencies.
+yum -y install autoreconf automake autoconf libtool openssl openssl-devel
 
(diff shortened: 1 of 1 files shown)


timiblossom:
@redboy1972 that diagram is nice and it helps new users to understand Dynomite quickly.  However, it is not the exact version that we have or intent to have.  I suggest to commit it somewhere into our docs directory but not on the front page.  Thanks.

ipapapa:
@redboy1972 We do not commit any more to the master branch. Might you please consider the changes proposed @timiblossom and create a pull request for Netflix:dev? I will temporarily close this issue.

PR #126, opened 2015-05-04
Reduce compilation warnings

shailesh33:
Following changes:
o Most of the changes in the patch are cosmetic, so basic as replacing tabs with spaces.
o minor bug in usage of sizeof.
o Change position in dyn_test.c from char * to size_t.
Minimal impact since changes are mostly in dyn_test.c

Testing:
ran dyn_test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,6 @@ dynomite
 Makefile
 Makefile.in
 /Debug
+/src/test
+cscope.out
+/src/cscope.out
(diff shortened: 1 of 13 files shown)


timiblossom:
Shailesh,

Can you please separate these 2 issues to make them clearer?  Also these need to go to Netflix's dev branch.  Thanks.

shailesh33:
Hi
Are you talking about compilation warning and cluster describe? If so, i have reverted the cluster describe patch. 

-SB

PR #133, opened 2015-05-20
Put in a check so we can avoid freeing a null message in the case we …

timiblossom:
…hit the memory limit


shailesh33:
what kind of message could this be?

timiblossom:
These are the incoming and outgoing messages in which we need to allocate memory for each one of them.

PR #151, opened 2015-08-10
PATCH IV For consistency

shailesh33:


timiblossom:
LGTM in response handling.  However, we should optimize more by not copying objects during fanning out by introducing more read pointers to msg buffers.  Also, change the hardcoded value of 3 for #replicas to be interpreted from the layout topology.

shailesh33:
Not copying objects?  I will prefer to make that a separate patch later on.
The max value is hardcoded right now. But we still get the actual value from the rack information.

PR #153, opened 2015-08-19
Rectify Dnode msg id

shailesh33:
While testing the consistency feature, it was observed that there were a lot of
inconsistencies. Further digging it was found out that the dnode message responses
were off by one. This patch is for the ground work for that issue.

This patch uses the request's message id as the id in the dnode msg header. That
way the responses and requests can be matched.
There was another bug where while sending responses to the dnode message, the code
would use the same dnode msg id for all the responses (that was the dnode msg id
for the dnode msg at the head of the out queue). Rectified that too. Yay!!


timiblossom:
The id assignment looks good.  However, we still need the patch to actually fix the issue that leverage the msg id for comparison and matching between request/response pair.

PR #155, opened 2015-08-21
set_consistency interface on stats port

shailesh33:
o to keep things going faster implement a stats port command to set consistency
    o seperate url for read and write
    o tested using : curl http://localhost:22224/set_consistency/write/local_quorum
      and curl http://localhost:22224/set_consistency/write/local_one
    o Important Caveat: With this all messages will start having the same
      consistency level irrespective of the connection.


timiblossom:
LGTM

PR #156, opened 2015-08-21
Rectify the dnode request response mismatch

shailesh33:
o This patch skips over the dnode request in the connection outqueue if it
      sees a response from a latter request.
    o Adds some extra debugging just in case there are issues.
    o Adds extra logging when non consistent data is seen
    o Rename variables to exactly depict what is req and what is rsp


timiblossom:
Look great

PR #175, opened 2015-10-14
Quorum Consistency Optimization

shailesh33:
These patches do the following:
o responds earlier to the requests as soon as quorum is achieved instead of waiting for all responses.
o Solves a bunch of crashes related to early closing of sockets, timeouts etc. These show up in quorum consistency
o Add ASSERT_LOG : here we can add extra information in case the assert fails.
o Logs will now show milliseconds. This can be helpful in debugging.
o Adds good comments on the new variables added

This patch could be the first step towards zero message copy in dynomite.

### DIFF
--- a/src/dyn_client.c
+++ b/src/dyn_client.c
@@ -23,7 +23,7 @@
 /**
  * This is the client connection. It receives requests from the client, and
  * forwards it to the corresponding peers and local data store server (if this
- * node owns the token.
+ * node owns the token).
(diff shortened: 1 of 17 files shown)


ipapapa:
Let us discuss offline if we need to update the configure.ac with the new version as well and what number to use, 0.5.6? Secondly, since these are patches that fix bugs in the connections for 0.5.4, I propose to create a new release once merged to master.

PR #185, opened 2015-10-26
Fix the crash when gossip is working

wuyingfengsui:
Fix the issue https://github.com/Netflix/dynomite/issues/182


shailesh33:
I think the problem is something else. Even not the assertion. I think the connection c_conn is getting reused and we are trying to access it. We would never expect c_conn & conn to be both of type PEER_SERVER. I will dig more into this.

ipapapa:
has been fixed with #186. To be merged with master in 0.5.6

PR #201, opened 2015-11-30
exponential backoff in removing fragmentation script

ipapapa:
Updated the script to add bigger sleep time when waiting for background AOF rewrite. Effectively, it checks if the flushing to disk has finished and adds 5 seconds to the previous time. The process will quit if background AOF takes more than 30 minutes.

### DIFF
--- a/scripts/dyno_redis_bgrewriteaof.sh
+++ b/scripts/dyno_redis_bgrewriteaof.sh
@@ -32,7 +32,7 @@ function fatal() { error "$*"; exit 1; }
 function usage_fatal() { error "$*"; usage >&2; exit 1; }
 
 # Default values for threshold for available node memory in KB
-THRESHOLD_MEMORY=5000000
+FREE_MIN_MEMORY=5000000
 # Default value for threshold for Redis RSS framgentation
 THRESHOLD_REDIS_RSS=1.5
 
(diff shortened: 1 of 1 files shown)


scripts/dyno_redis_bgrewriteaof.sh
```diff
@@ -128,7 +144,11 @@ if [[ ${FREE_MEMORY} -le ${THRESHOLD_MEMORY} ]]; then
        log "INFO: Redis RSS fragmentation is $REDIS_RSS_RAM < $THRESHOLD_REDIS_RSS . Exiting..."
      fi
 else
-    log "INFO: Available memory is $(($FREE_MEMORY/1000))  more than $(($THRESHOLD_MEMORY/1000)) KB. Exiting..."
+    if [[ ${FREE_MIN_MEMORY} -le 1000 ]]; then
+       log "INFO: Available memory is $FREE_MEMORY KB more than $FREE_MIN_MEMORY KB. Exiting..."
+    else
+       log "INFO: Available memory is $(($FREE_MEMORY/1000)) MB more than $(($FREE_MIN_MEMORY/1000)) MB. Exiting..."
```
shailesh33: this should be /1024?

PR #207, opened 2015-12-09
Added Florida OS_ENV_VARS to allow multi-cluster solution

diegopacheco:
@timiblossom  @ipapapa 

Sorry for make this in one PR but i end up figuring this out just later. On the previous PR i add the ability to pass CFLAG with Florida IP, PORT and REQUEST to dynomite. However this requires to keep building dynomite all the time, in my case i want have multiple clusters so i need have different Florida URLs providing different seeds for each cluster. Make dynomite all the time makes the build process slow, so i change the code to look for ENV var like:
  - DYNOMITE_FLORIDA_IP
  - DYNOMITE_FLORIDA_PORT
  - DYNOMITE_FLRORIDA_REQUEST
I made this change be processed just once, so will read the ENV_VARS just one time.
IF this OS_VARS are not present(NULL) the code will use the macros or CFLAGS values.
This is great because allow to use the same binary all the time and the later when i deploy the code into ASGARD i can have just different user data since is just a matter of export different values to the AMI/Linux. There is a minor FIX of YAML syntax as well. 

So for instance someone can have on the /etc/profile 
```bash

export DYNOMITE_FLORIDA_PORT=8080
export DYNOMITE_FLORIDA_IP="127.0.0.1"
export DYNOMITE_FLORIDA_REQUEST="GET /florida/cluster1/get_seeds.txt HTTP/1.0
Host: 127.0.0.1
User-Agent: HTMLGET 1.0

"
```
and for another box just
```bash

export DYNOMITE_FLORIDA_PORT=8080
export DYNOMITE_FLORIDA_IP="127.0.0.1"
export DYNOMITE_FLORIDA_REQUEST="GET /florida/cluster2/get_seeds.txt HTTP/1.0
Host: 127.0.0.1
User-Agent: HTMLGET 1.0

"
```

Even pass this through ASGARD via user_data. 

Them build dynomite normaly like:
```bash
sudo autoreconf -fvi ; sudo ./configure --enable-debug=log ; sudo make;
```
And Run:
```bash
 sudo --preserve-env src/dynomite -c conf/dynomite_florida_single.yml 
```
And it works :-) 

Cheers,
Diego Pacheco


ipapapa:
@diegopacheco May you please also add a description under the docs folder on how one can configure the profile and then run with the --preserve-env flag to run Florida? (pretty much what you have here) It would assist in the reusability of this code path. What do you think?

diegopacheco:
@ipapapa Sure. I will do it.

ipapapa:
@diegopacheco thank you, I will keep this PR open for your doc changes.

diegopacheco:
@ipapapa I just added some doc as you requested.

PR #208, opened 2015-12-10
Do not replicate INFO to other nodes for quorum

shailesh33:


ipapapa:
Good catch, INFO should not be propagated to other nodes.

PR #213, opened 2015-12-14
configure.ac: stray merge artifact

proger:
Lol'd it has zero effect on the build process :)

PR #214, opened 2015-12-14
Removed garbage

ip1981:


ipapapa:
@ip1981 thanks for taking a look at it. We are following Gitflow so we do not commit to master anymore... The next release coming soon https://github.com/Netflix/dynomite/blob/release/v0.5.6/configure.ac will have that fixed.

ipapapa:
This seems to be the same as #213. I am closing this, and will keep the #213 since it came first open for a couple of days in case there are any further comments.

PR #222, opened 2015-12-23
dyn_util: use /proc/$pid/exe to detect current running executable

proger:
while here, redirect addr2line to stderr to stay consistent with log output

### DIFF
--- a/src/dyn_util.c
+++ b/src/dyn_util.c
@@ -320,10 +320,10 @@ dn_stacktrace(int skip_count)
         loga("[%d] %s", j, symbols[i]);
 
         char syscom[256];
-         sprintf(syscom,"addr2line %p -e src/dynomite", stack[i]); //last parameter is the name of this app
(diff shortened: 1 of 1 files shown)


ipapapa:
Looks OK as a logic. Did you tried it out in your stacktrace? Moreover, I was wondering if does make sense to leverage (pid)[https://github.com/Netflix/dynomite/blob/dev/src/dyn_core.h#L194] but could not determine from a very quick look a way pass the struct instance so it sounds fine.

proger:
Yep, works just fine in my setup (though I had to make binutils available ofc).

`getpid()` return is cached by glibc anyway, so i wouldn't worry.

FWIW it's also possible to `readlink("/proc/self/exe", buf, len)` and then use `buf` as an argument to `snprintf` but ¯\_(ツ)_/¯

proger:
\o/

PR #233, opened 2016-01-08
Suppress logging if a request cannot have a quorum response.

timiblossom:


src/dyn_response_mgr.c
```diff
@@ -125,8 +125,12 @@ rspmgr_get_response(struct response_mgr *rspmgr)
...
     }
     rspmgr_incr_non_quorum_responses_stats(rspmgr);
-    log_error("none of the responses match, returning first");
-    if (log_loggable(LOG_ERR)) {
+
+    if (log_loggable(LOG_NOTICE)) {
+      log_error("none of the responses match, returning first");
+    }
+
+    if (log_loggable(LOG_VVERB)) {
```
shailesh33: I had not expected this to be so verbose. Sorry about that. but can you make this one as LOG_INFO?

PR #235, opened 2016-01-08
Basic functionality test framework

shailesh33:
o connects to an existing cluster (hardcoded)
    o does a dual write to redis and dynomite and raises
      assertion if results differ
    o runs basic key value, large payload and hash map test
    o connects to a random node in the cluster
    o Integrates with travis


conf/a_dc2_rack1_node1.yml
```diff
@@ -1,13 +1,13 @@
...
+  dyn_listen: 127.0.0.2:8101
   dyn_seeds:
-  - 127.0.0.1:8101:rack1:dc1:12345678
-  - 127.0.0.2:8101:rack2:dc2:1383429731
-  - 127.0.0.3:8101:rack2:dc2:12345678
-  - 127.0.0.5:8101:rack1:dc2:12345678 
-  listen: 127.0.0.4:8102
+  - 127.0.0.1:8101:rack1:dc1:1383429731
+  - 127.0.0.4:8101:rack2:dc2:1383429731
+  - 127.0.0.5:8101:rack2:dc2:12345678
```
ipapapa: Do we want to have half the size of the token?

test/func_test.py
```diff
@@ -0,0 +1,142 @@
...
+    # finally do a hscan
+    #key, _ = create_key_field()
+    #next_index = 0;
+    #while True:
+        #result = c.run_verify("hscan", key, next_index)
+        #next_index = result[0] 
+        #print next_index
+        #if next_index == 0:
+            #break
+
```
ipapapa: Add comments about how you do the port distribution :)

conf/a_dc2_rack1_node1.yml
shailesh33: thats the way existing yml files are.

Issue #206, opened 2015-12-06
Update scripts/dynomite.init to be LSB compliant


akbarahmed:
Is there any interest in updating the `scripts/dynomite.init` script to be LSB compliant?

I've started to update this script in my forked repo. I'll issue a pull request when I'm done if this is a desired feature.

shailesh33:
Sure go ahead. more helping hands are always welcome

ipapapa:
@akbarahmed It is indeed great; please consider issuing the pull request to the "dev" branch.

akbarahmed:
Quick status update. I'm almost done with a 1st draft of the LSB compliant init script.

Issue #228, opened 2016-01-06
ForestDB integration


timiblossom:
There are use cases that we have data set bigger than the available physical memory.  This means we need to store hot data in memory and cold data on disk, and then have an efficient mechanism to do CRUB operations.   ForestDB seems to be a good candidate for this purpose.

ipapapa:
We have created a project for this issue, with a card on ForestDB.

PR #236, opened 2016-01-12
Ubuntu lsb upstart

akbarahmed:
close Netflix/dynomite#206

Initial draft of an LSB-compliant init script for Ubuntu Upstart. 

In terms of pending tasks, I need to change the init script to run the `dynomite` daemon as the `dynomite` user.

I also need to write several additional scripts including one for RHEL <= 6.7, RHEL > 6.7 for systemd, and one for Ubuntu systemd.

Beyond the above, the init scripts need to be added to the build / packaging flow so that we can build .rpm and .deb packages.


init/dynomite_ubuntu_upstart
```diff
@@ -0,0 +1,189 @@
...
+# Default-Start:     2 3 4 5
+# Default-Stop:      0 1 6
+# Short-Description: dynamo layer for different k-v storage engines
+# Description:       dynomite is a thin dynamo proxy layer that provides
+#                    high availability (HA) and multi-datacenter replication
+#                    for Redis and MemCached.
+### END INIT INFO
+
+NAME=dynomite
+DESC="distributed cache layer"
```
ipapapa: I would call it distributed data store layer. Right now there is support for Redis, but other persistent data stores may be supported in the near future. See #228

init/dynomite_ubuntu_upstart
```diff
@@ -0,0 +1,189 @@
...
+#
+# Return via echo
+# true = PIDFILE exists and process is running
+# false = PIDFILE exists but process is not running
+# false = PIDFILE does not exist
+#
+is_running () {
+    if [ -s $PIDFILE ] ; then
+        PID=`head -n 1 $PIDFILE 2>/dev/null`;
+        if [ ! -z "`ps --no-heading $PID`" ] ; then
```
ipapapa: I am not sure if you can change that, but we have seen cases for some reason a port is hanging because of OS bug, or other reasons, that Dynomite may be seen up but would not have started properly. So if this is targeted to Redis users the safe way to go is to issue a "redis-cli -p 8102 ping" and grep for PONG. Similarly for memcached etc.

init/dynomite_ubuntu_upstart
akbarahmed: Done and pushed.

init/dynomite_ubuntu_upstart
akbarahmed: @ipapapa I removed `redis` from `Required-Start` and `Required-Stop` for now. I have pushed this change a few minutes ago.

To use Upstart for maximum effect I should write an init script generator where the user is asked which backends they want to support. This way `Required-Start` and `Required-Stop` can be changed based on the specific backends used on each installation. Specifically, `Required-Start` and `Required-Stop` order the init scripts so that `dynomite` is started after the backend(s).

With regard to checking if the Redis or memcached process is alive, I don't think that should be a part of the init script for Dynomite (for now). However, after writing an init script generator, it may make sense to customize `is_running()` to also check / test for dynomite + a specific backend.

init/dynomite_ubuntu_upstart
```diff
@@ -0,0 +1,189 @@
...
+# Description:       dynomite is a thin dynamo proxy layer that provides
+#                    high availability (HA) and multi-datacenter replication
+#                    for Redis and MemCached.
+### END INIT INFO
+
+NAME=dynomite
+DESC="distributed dynamo layer"
+PATH=/sbin:/usr/sbin:/usr/local/sbin:/bin:/usr/bin
+PIDFILE=/var/run/${NAME}.pid
+CONFIG=/etc/dynomite/${NAME}.yml
```
ipapapa: All .yml files are under dynomite/conf/. Shouldn't this point to this sub-folder?

init/dynomite_ubuntu_upstart
```diff
@@ -0,0 +1,189 @@
...
+
+#
+# Start dynomite
+#
+# Return
+# 0 = daemon has been started
+# 1 = daemon was already running
+# 2 = daemon could not be started
+#
+start_dynomite () {
```
ipapapa: start_dynomite and stop_dynomite already exists under https://github.com/Netflix/dynomite/tree/dev/bin in scripts Shouldn't we call them as functions instead? What do you think?

akbarahmed: @ipapapa Yes, I saw those before writing this init script. However, I don't think they are what we want to use for an LSB compliant init script that closely resembles other init scripts.

In this init script we use `start-stop-daemon` to start and stop the daemon because its the standard way of starting system processes on Ubuntu.

On RHEL, I'll use the `daemon` function which is the standard way to start a background process on RHEL/Centos.

While developing this init script I used NGINX, PostgreSQL, Sendmail, and Bind9 as references. I referred to the init scripts installed via `apt-get` (the init scripts created by the package maintainers) and not the init scripts included in the upstream source.

Please let me know if you think using `start-stop-daemon` is appropriate.

ipapapa:
@akbarahmed I had one more comment on the conf location not sure if you saw that above, or it was missed. Please let me know.

init/dynomite_ubuntu_upstart
akbarahmed: @ipapapa These are good questions. I'll list the questions below as a numbered list, then answer each one.

1. What should be the $CONFIG directory?
2. Does a single config file limit an instance of Dynomite to one deployment?
3. Should the user be able to override the default config file and specify a custom one?

### Replies

#### 1 The CONFIG directory

I set `/etc/dynomite` as the config directory in the init script because `/etc/service-name/` is a standard location for configuration files. My assumption with the init script is that it will be part of a package or other installation script or `make install` that moves files from the build directory to the install location in the file system.

IMO, the init script is only useful after `make install` (or eventually `apt-get install` or `yum install` once we have build packages) is run.

#### 2 Single config file

I think there are a few ways that we can handle this. Per my understanding, the `dynomite` binary only accepts one configuration file via `-c` when it's started. Namely there is only one config file used per dynomite process and its normal to have only one instance of dynomite per server (i.e. there is a single dynomite binary per server). 

#### 3 Override of default config file 

There are a few possible solutions to allow users to change which conf file is used by `dynomite`. 

a. The user could copy the desired `.yml` file to `/etc/dynomite/dynomite.yml`. i.e. `cp conf/redis_dc1.yml /etc/dynomite/dynomite.yml`. This can be automated via tooling, but the concept is a basic copy.
b. The installer could copy all of the files in `conf/*` to `/etc/dynomite/conf`. Then we could create an `enabled` directory such as `/etc/dynomite/enabled`. Then, the user could create a symlink to the desired conf file in `enabled`. This would be similar to how the `default.d` directory works for nginx. I'd have to update the path to support this solution, but its a 10 second update.
c. We could allow the user to symlink from `conf/some-file.yml` to `/etc/dynomite/dynomite.yml`. This would be similar to how `systemd` works.
d. Any of the above values can be overridden in `/etc/default/dynomite`. I allow this option via the `INIT_OPTIONS` variable in the script. This is the standard way of overriding values set in an init script. Further, this method allows user customizations to persist after a package upgrade.

I don't yet have an opinion on which of the options (1, 2 or 3) above is ideal for Dynomite. It's something I'm actively thinking about.

ipapapa: IMHO leaving it to default in dynomite.yml is OK. We can improve it later so that {NAME} is not called in both for .pid for .yml, so if somebody wants to change one it does not affect both.

PR #239, opened 2016-01-21
Adding a timeout factor to help investigate

shailesh33:
This change will add a factor to the timeouts. This will help investigate
and resolve some of the issues that we occasionally see.
The factor is between 1 & 4 (integer)


ipapapa:
Looks good

PR #273, opened 2016-03-20
Standardize binary names, increment version

akbarahmed:
I noticed that the current binary names do not have a consistent pattern (i.e. `dynomite`, `test`, `dyno_hash_tool`). This pull request creates a standard naming convention of `dynomite-` for server-related binaries in the `sbin` directory, and `dyno-` for client-related binaries in the `bin` directory.

With the new names, a user can type `dyno` + `tab tab` and view all Dynomite related commands.

- Increment version in `configure.am` (was 0.5.6, now 0.5.7)
- Rename binaries to naming standard/convention
    - `dynomite-` = server-related binaries
    - `dyno-` = client-related binaries
    - Use '-' not '_' as easier/faster for admins to type (i.e. better UX). SysAdmins type commands repeatedly and dash is a single key whereas underscore is two keys.
- Rename binary `test` to `dynomite-test`
- Rename binary `dyn_hash_tool` to `dyno-hash-tool`
- Move `dyno-hash-tool` to `bin` directory
- Update help message for `dyno-hash-tool`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# IDE
+.idea
+
 # Compiled Object files
 *.lo
 *.o
@@ -37,6 +40,8 @@ TODO
 core*
 nutcracker
 dynomite
+dynomite-test
+dyno-hash-tool
 
 # extracted yaml
 !/contrib/yaml-0.1.4.tar.gz
--- a/configure.ac
+++ b/configure.ac
@@ -5,7 +5,7 @@ m4_ifndef([m4_esyscmd_s], [m4_define([m4_esyscmd_s], [m4_chomp_all(m4_esyscmd([$
 # Define the package version numbers and the bug reporting address
 m4_define([DN_MAJOR], 0)
 m4_define([DN_MINOR], 5)
-m4_define([DN_PATCH], 6)
+m4_define([DN_PATCH], 7)
 m4_define([DN_BUGS], [dynomite@netflix.com])
 m4_define([DN_VERSION_STRING], m4_esyscmd_s([git describe --dirty --always --tags]))
 
--- a/configure.ac
+++ b/configure.ac
@@ -2,10 +2,7 @@
 m4_ifndef([m4_esyscmd_s], [m4_define([m4_chomp_all], [m4_format([[%.*s]], m4_bregexp(m4_translit([[$1]], [/], [/ ]), [/*$]), [$1])])])
 m4_ifndef([m4_esyscmd_s], [m4_define([m4_esyscmd_s], [m4_chomp_all(m4_esyscmd([$1]))])])
 
-# Define the package version numbers and the bug reporting address
-m4_define([DN_MAJOR], 0)
-m4_define([DN_MINOR], 5)
-m4_define([DN_PATCH], 7)
+# Define the bug reporting address and package version
 m4_define([DN_BUGS], [dynomite@netflix.com])
 m4_define([DN_VERSION_STRING], m4_esyscmd_s([git describe --dirty --always --tags]))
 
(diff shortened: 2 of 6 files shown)


configure.ac
```diff
@@ -5,7 +5,7 @@ m4_ifndef([m4_esyscmd_s], [m4_define([m4_esyscmd_s], [m4_chomp_all(m4_esyscmd([$
 # Define the package version numbers and the bug reporting address
 m4_define([DN_MAJOR], 0)
 m4_define([DN_MINOR], 5)
-m4_define([DN_PATCH], 6)
+m4_define([DN_PATCH], 7)
```
shailesh33: is this patch being used anywhere? this was tedious so I moved to use git version for the binary version.

akbarahmed: @shailesh33 The following variables are not used anywhere:
- `DN_MAJOR`
- `DN_MINOR`
- `DN_PATCH`

I have deleted these variables from the `configure.ac` file which has updated the PR.

src/tools/Makefile.am
```diff
@@ -23,9 +23,9 @@ if OS_SOLARIS
 AM_LDFLAGS += -lnsl -lsocket
 endif
 
-sbin_PROGRAMS = dyn_hash_tool
+bin_PROGRAMS = dyno-hash-tool
```
shailesh33: In that case this should be dynomite-hash-tool.
This can be optional in the build process.

.gitignore
```diff
@@ -37,6 +40,8 @@ TODO
 core*
 nutcracker
 dynomite
+dynomite-test
+dyno-hash-tool
```
shailesh33: dynomite-hash-tool

src/tools/Makefile.am
akbarahmed: @shailesh33 My earlier comment about using `dyno-` for all non-system binaries and `dynomite-` for system binaries was wrong. This rule falls apart with any CLI tools where dyno is a better prefix, but where the tool modifies things like conf files, etc (which would put the `dyno-` prefix tool in `sbin`).

That said, what do you think about using `dyno` as the prefix for tools, utilities, etc. that a user or admin would execute repeatedly?

At this point I don't have a general rule for how naming should work other than to think that a short prefix like `dyno` is better for any command that is called often.

shailesh33: If you ask me, I see files named dyn_ in the source code. I has named dyn_hash_tool based on that. dyno- would create confusion with dyno client.

akbarahmed: @shailesh33 IMO, the `dyn_` source code file names are Dynomite developer focused, while the `dyno-` and `dynomite-` command names are user and sysadmin focused.

Most of our users will never write C source. In this way, I think the naming in one context should not influence the other.

With regard to the `dyno` client driver, I thought about this as well. This is a 3rd context, that of the application developer who writes apps that use Dynomite. Using the `dyno` library within a Maven or Gradle file is distinct from `dyno` or `dynomite` CLI tools and binaries.

With that said, completely different names would be ideal. However, I think naming the `dyno` library as `dynolib` may be a bit of overkill. To elaborate, I would expect that the client libraries in each language have the same name of `dyno`. So we'd see something like `require dyno` in Node.js and `import dyno` in Go, etc.

Per my thinking, the following are a few of the district roles that use Dynomite, each of which may have naming standards appropriate to their environment.

- Dynomite developer: A person who works on Dynomite's source
- User/SysAdmin: A person who installs, configures, operates Dynomite, Redis and related services
- Application developer: A person who developers applications in Java, Go, Node.js/JS, etc. and uses a client driver and the Redis API to query Dynomite

.gitignore
akbarahmed: @shailesh33 I have renamed the binary to `dynomite-hash-tool` per your request.

PR #276, opened 2016-03-29
Fix a bunch of compiler warnings

shailesh33:
o Defines new types for usec_t and msec_t
    o comment out unused functions
    o type conversion errors

### DIFF
--- a/src/dyn_conf.c
+++ b/src/dyn_conf.c
@@ -97,7 +97,7 @@ static struct command conf_commands[] = {
 
     { string("server_retry_timeout"),
       conf_set_num,
-      offsetof(struct conf_pool, server_retry_timeout) },
+      offsetof(struct conf_pool, server_retry_timeout_ms) },
 
     { string("server_failure_limit"),
       conf_set_num,
@@ -340,7 +340,7 @@ conf_pool_init(struct conf_pool *cp, struct string *name)
     cp->preconnect = CONF_UNSET_NUM;
     cp->auto_eject_hosts = CONF_UNSET_NUM;
     cp->server_connections = CONF_UNSET_NUM;
-    cp->server_retry_timeout = CONF_UNSET_NUM;
+    cp->server_retry_timeout_ms = CONF_UNSET_NUM;
     cp->server_failure_limit = CONF_UNSET_NUM;
 
     //initialization for dynomite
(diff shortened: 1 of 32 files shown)


src/dyn_conf.c
```diff
@@ -485,7 +485,7 @@ conf_pool_each_transform(void *elem, void *data)
     sp->client_connections = (uint32_t)cp->client_connections;
 
     sp->server_connections = (uint32_t)cp->server_connections;
-    sp->server_retry_timeout = (int64_t)cp->server_retry_timeout * 1000LL;
+    sp->server_retry_timeout_us = (int64_t)cp->server_retry_timeout_ms * 1000LL;
```
ipapapa: is there a reason to keep both "us" and "ms" in different structures? If we agree in one unit that is important for our platform we can avoid the multiplications *1000LL.

src/dyn_connection.c
```diff
@@ -191,8 +207,8 @@ _conn_get(void)
     conn_set_write_consistency(conn, g_write_consistency);
     conn->type = CONN_UNSPECIFIED;
 
-    unsigned char *ase_key = generate_aes_key();
-    strncpy(conn->aes_key, ase_key, strlen(ase_key)); //generate a new key for each connection
+    unsigned char *aes_key = generate_aes_key();
+    strncpy((char *)conn->aes_key, (char *)aes_key, strlen((char *)aes_key)); //generate a new key for each connection
```
ipapapa: aes_key is used only as an unsigned char in the dyn_crypto code. More specifically in the openssl API: https://www.openssl.org/docs/manmaster/crypto/EVP_EncryptInit.html the key/iv are provided as unsigned chars. I feel there is no reason for us to keep the conn->aes_key as char. This will avoid the complexities of typecasting it. Can we change the conn->aes_key instead of typecasting?

ipapapa:
Have you done any testing on these, or this will come once merge to dev?

shailesh33:
I am creating an ami for it.

src/dyn_conf.c
shailesh33: no reason. Will make the change

src/dyn_connection.c
shailesh33: I am typecasting it for strncpy. dyn_connection has unsigned char just like dyn_crypto

shailesh33:
closing this pull request and recreating it on a different branch

src/dyn_connection.c
ipapapa: OK, thanks for explaining.

PR #281, opened 2016-04-04
Hotfix/ooo

shailesh33:


src/dyn_dnode_peer.c
```diff
@@ -1901,3 +1901,62 @@ init_dnode_peer_conn(struct conn *conn)
...
+rack_name_cmp(const void *t1, const void *t2)
+{
+    const struct rack *s1 = t1, *s2 = t2;
+
+    return string_compare(s1->name, s2->name);
+}
+
+// The idea here is to have a designated rack in each remote region to replicate
+// data to. This is used to replicate writes to remote regions
+static void
```
timiblossom: This approach could cause some disadvantages:
1- By picking a remote rack based on itself's rack index and due to non-sorting on the rack names, the same remote rack could be chosen by multiple nodes on different DCs and Racks.  This could cause some rack to have extra loads in a DC than the rest of the racks in the same DC.

2- We can not deploy dynomite in  an asymmetric way.  After this PR, we are asking all DCs to have the same number of racks and this is fundamentally changing the Dynomite architecture

3- We won't be able to take down a whole rack in a DC for maintenance if that rack is the designated rack for others to send the requests.  Today, we can do this without any issue.

src/dyn_dnode_peer.c
```diff
@@ -1901,3 +1901,62 @@ init_dnode_peer_conn(struct conn *conn)
...
+        if (string_compare(dc->name, &sp->dc) != 0)
+            continue;
+
+        // if the dc is a local dc, get the rack_idx
+        uint32_t rack_index;
+        uint32_t rack_cnt = array_n(&dc->racks);
+        for(rack_index = 0; rack_index < rack_cnt; rack_index++) {
+            struct rack *rack = array_get(&dc->racks, rack_index);
+            if (string_compare(rack->name, &sp->rack) == 0) {
+                my_rack_index = rack_index;
```
timiblossom: Need to sort or same rack_index could be used by multiple nodes

shailesh33:
1. thats the reason we sort the rack names before anything.
2. It does not. we do a modulo, its a best effort way of doing things.
3. We cannot do that today. We try to reroute the message to a different node in the same rack only. There is a TODO for it that needs to be fixed.

src/dyn_dnode_peer.c
shailesh33: arent we sorting it on line 1925

src/dyn_dnode_peer.c
shailesh33: 1. thats the reason we sort the rack names before anything.
2. It does not. we do a modulo, its a best effort way of doing things.
3. We cannot do that today. We try to reroute the message to a different node in the same rack only. There is a TODO for it that needs to be fixed.

timiblossom: 1. Yeah, you're right.  I missed that as the method is too long :)
2. Yep, after this PR.  Every DC has to have the same number of racks.  Otherwise, I suggest that we should hash keys and if they belong to the same bucket, route them to the same rack or node.
3. Today, we can shutdown an entire rack for a binary upgrading or other maintenance operations such as restarting Redis/instances.

PR #282, opened 2016-04-04
Remove round robin replication to remote regions

shailesh33:
If you do round robin for every message to remote region, you end up the
messages reaching remote nodes in out of order since some can have extra hops.
Fix it by having a designated remote rack to send messages to.
Every rack in a local region have a designated remote region rack.

### DIFF
--- a/src/dyn_client.c
+++ b/src/dyn_client.c
@@ -787,8 +787,9 @@ req_forward_remote_dc(struct context *ctx, struct conn *c_conn, struct msg *msg,
     if (rack_cnt == 0)
         return;
 
-    uint32_t ran_index = (uint32_t)rand() % rack_cnt;
-    struct rack *rack = array_get(&dc->racks, ran_index);
+    struct rack *rack = dc->preselected_rack_for_replication;
+    if (rack == NULL)
+        rack = array_get(&dc->racks, 0);
 
     struct msg *rack_msg = msg_get(c_conn, msg->request, msg->data_store, __FUNCTION__);
     if (rack_msg == NULL) {
--- a/src/dyn_core.c
+++ b/src/dyn_core.c
@@ -192,6 +192,7 @@ core_ctx_create(struct instance *nci)
 	CBUF_Init(C2S_OutQ);
 
 	gossip_pool_init(ctx);
+    preselect_remote_rack_for_replication(ctx);
 
 	log_debug(LOG_VVERB, "created ctx %p id %"PRIu32"", ctx, ctx->id);
 
--- a/src/dyn_core.h
+++ b/src/dyn_core.h
@@ -215,6 +215,7 @@ struct rack {
 struct datacenter {
 	struct string      *name;            /* datacenter name */
 	struct array       racks;           /* list of racks in a datacenter */
+    struct rack        *preselected_rack_for_replication;
 	dict               *dict_rack;
 };
 
--- a/src/dyn_dnode_peer.c
+++ b/src/dyn_dnode_peer.c
@@ -1941,14 +1941,23 @@ preselect_remote_rack_for_replication_each(void *elem, void *data)
 
     for(dc_index = 0; dc_index < dc_cnt; dc_index++) {
         struct datacenter *dc = array_get(&sp->datacenters, dc_index);
+        dc->preselected_rack_for_replication = NULL;
+
+        // Nothing to do for local DC, continue;
         if (string_compare(dc->name, &sp->dc) == 0)
             continue;
+
+        // if no racks, keep preselected_rack_for_replication as NULL
+        uint32_t rack_cnt = array_n(&dc->racks);
+        if (rack_cnt == 0)
+            continue;
+
         // if the dc is a remote dc, get the rack at rack_idx
         // use that as preselected rack for replication
-        uint32_t rack_cnt = array_n(&dc->racks);
         uint32_t this_rack_index = my_rack_index % rack_cnt;
-        dc->preselected_rack_for_replication = array_get(&dc->racks, this_rack_index);
-        log_notice("Selected rack %.*s for remote region %.*s",
+        dc->preselected_rack_for_replication = array_get(&dc->racks,
+                                                         this_rack_index);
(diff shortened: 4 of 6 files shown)


src/dyn_core.c
```diff
@@ -192,6 +192,7 @@ core_ctx_create(struct instance *nci)
 	CBUF_Init(C2S_OutQ);
 
 	gossip_pool_init(ctx);
+    preselect_remote_rack_for_replication(ctx);
```
ipapapa: Is the indentation correct?

src/dyn_dnode_peer.c
```diff
@@ -1901,3 +1901,62 @@ init_dnode_peer_conn(struct conn *conn)
...
+rack_name_cmp(const void *t1, const void *t2)
+{
+    const struct rack *s1 = t1, *s2 = t2;
+
+    return string_compare(s1->name, s2->name);
+}
+
+// The idea here is to have a designated rack in each remote region to replicate
+// data to. This is used to replicate writes to remote regions
+static void
```
ipapapa: Is the code so safe that the function will never return DN_ERROR? aka rstatus_t

shailesh33: refactor? how?

src/dyn_dnode_peer.c
```diff
@@ -1901,3 +1901,62 @@ init_dnode_peer_conn(struct conn *conn)
...
+    }
+
+    for(dc_index = 0; dc_index < dc_cnt; dc_index++) {
+        struct datacenter *dc = array_get(&sp->datacenters, dc_index);
+        if (string_compare(dc->name, &sp->dc) == 0)
+            continue;
+        // if the dc is a remote dc, get the rack at rack_idx
+        // use that as preselected rack for replication
+        uint32_t rack_cnt = array_n(&dc->racks);
+        uint32_t this_rack_index = my_rack_index % rack_cnt;
```
ipapapa: can rack_cnt ever be zero? If no, then in dyno_client why do we check if it is zero before the modulo? If yes, shall return in case it is zero?

src/dyn_dnode_peer.c
ipapapa: mostly return rstatus_t here. THROW_STATUS in dyn_server.c and deinit the variables in dyn_core.c.

src/dyn_dnode_peer.c
shailesh33: are you talking about check for NULL in dyn_client.c? thats a catch_all just in case there is something that we never processed.

src/dyn_core.c
shailesh33: yes, for the preselected . The rest of the code is out of place

src/dyn_dnode_peer.c
ipapapa: This is what I am talking about: https://github.com/Netflix/dynomite/blob/dev/src/dyn_client.c#L787

timiblossom: I think Yannis is right.  There are normal cases that those 1st nodes won't have enough information about a remote rack.  Thus, the rack count would be zero in that case and we don't want them to crash

src/dyn_client.c
```diff
@@ -787,8 +787,9 @@ req_forward_remote_dc(struct context *ctx, struct conn *c_conn, struct msg *msg,
     if (rack_cnt == 0)
         return;
 
-    uint32_t ran_index = (uint32_t)rand() % rack_cnt;
-    struct rack *rack = array_get(&dc->racks, ran_index);
+    struct rack *rack = dc->preselected_rack_for_replication;
```
timiblossom: I don't see the logic where you handle the failure scenario where the whole rack goes down for maintenance or if a node in that rack goes down.  With the preselected_rack_for_replication only gets initialized one time, it is scary to touch the remote rack associated with this assignment as we might end up losing a lot of replicated data

src/dyn_dnode_peer.c
```diff
@@ -1901,3 +1901,62 @@ init_dnode_peer_conn(struct conn *conn)
...
+            struct rack *rack = array_get(&dc->racks, rack_index);
+            if (string_compare(rack->name, &sp->rack) == 0) {
+                my_rack_index = rack_index;
+                log_notice("my rack index %u", my_rack_index);
+                break;
+            }
+        }
+    }
+
+    for(dc_index = 0; dc_index < dc_cnt; dc_index++) {
```
timiblossom: In asymmetric deployment (3 nodes in us-east-1 and 2nodes in eu-west-1), a rack in eu-west-1 would receive more loads than to cause 99th latencies to be high which gives an impression that the cluster has a high 99th latencies

src/dyn_server.c
```diff
@@ -881,6 +881,7 @@ dc_init(struct datacenter *dc)
 	dc->dict_rack = dictCreate(&dc_string_dict_type, NULL);
 	dc->name = dn_alloc(sizeof(struct string));
 	string_init(dc->name);
+    dc->preselected_rack_for_replication = NULL;
```
timiblossom: Indentation?

src/dyn_core.h
```diff
@@ -215,6 +215,7 @@ struct rack {
 struct datacenter {
 	struct string      *name;            /* datacenter name */
 	struct array       racks;           /* list of racks in a datacenter */
+    struct rack        *preselected_rack_for_replication;
```
timiblossom: Indentation?

src/dyn_client.c
shailesh33: Thanks. We will fix this separately from a hotfix.

src/dyn_dnode_peer.c
shailesh33: Yes. Again, we will fix this separately from the hotfix

src/dyn_server.c
shailesh33: Its the rest of the code that has the issue. I will fix this across the entire code base seperately... replace tabs with 4 spaces

src/dyn_core.h
shailesh33: same as above

PR #285, opened 2016-04-07
adding support for CONFIG GET

ipapapa:
Adding support for CONFIG request operations that are not changing the Redis configuration.


src/dyn_message.h
```diff
@@ -183,6 +183,7 @@ typedef enum msg_type {
     MSG_RSP_REDIS_INTEGER,
     MSG_RSP_REDIS_BULK,
     MSG_RSP_REDIS_MULTIBULK,
+	MSG_REQ_REDIS_CONFIG,
```
shailesh33: I think this should be MSG_REQ_REDIS_CONFIG_GET. that way it is more specific.

src/proto/dyn_redis.c
```diff
@@ -1108,6 +1118,8 @@ redis_parse_req(struct msg *r)
                     break;
                 }
 
+                break;
```
shailesh33: good catch.

notes/redis.md
```diff
@@ -341,11 +341,13 @@
...
-    |    CONFIG GET     |    No      | CONFIG GET parameter                                                                                                |
+    |    CONFIG GET     |    YES     | CONFIG GET parameter                                                                                                |
     +-------------------+------------+---------------------------------------------------------------------------------------------------------------------+
     |    CONFIG SET     |    No      | CONFIG SET parameter value                                                                                          |
     +-------------------+------------+---------------------------------------------------------------------------------------------------------------------+
-    |  CONFIG RESETSTAT |    No      | CONFIG RESETSTAT                                                                                                    |
+    |  CONFIG RESETSTAT |    YES     | CONFIG RESETSTAT                                                                                                    |
```
shailesh33: where is this implemented?

src/dyn_message.h
ipapapa: I disagree with the this. When we do the string comparison we look for "CONFIG". Later we we remove all others, other than "GET".

notes/redis.md
ipapapa: fixed it, sorry

src/proto/dyn_redis.c
```diff
@@ -1393,7 +1406,14 @@ redis_parse_req(struct msg *r)
             break;
 
         case SW_ARG1:
+
+
+            if (r->type == MSG_REQ_REDIS_CONFIG && !str3icmp(m, 'g', 'e', 't')) {
```
shailesh33: shouldnt it be r->type == MSG_REQ_REDIS_CONFIG && str3icmp(m, 'g', 'e', 't')

you want to error out for anything other than get, and for that str3icmp would return non 0,

PR #291, opened 2016-05-05
Cleanup dyn_conf

shailesh33:
o Mainly try to keep dyn_conf.h clean so its understood whats the
interface
    o Create a new enum type of secure server option and use that
    o reduce the warnings in dyn_conf.c
    o Get ride of ctx_id

Sorry about the messy diff. Its only because of function reordering

### DIFF
--- a/src/dyn_client.c
+++ b/src/dyn_client.c
@@ -638,7 +638,7 @@ local_req_forward(struct context *ctx, struct conn *c_conn, struct msg *msg,
 
     conn_enqueue_inq(ctx, s_conn, msg);
     req_forward_stats(ctx, s_conn->owner, msg);
-    if(msg->data_store==DATA_REDIS){
+    if(g_data_store == DATA_REDIS){
     	req_redis_stats(ctx, s_conn->owner, msg);
 	}
 
@@ -731,8 +731,7 @@ req_forward_all_local_racks(struct context *ctx, struct conn *c_conn,
         if (string_compare(rack->name, &pool->rack) == 0 ) {
             rack_msg = msg;
         } else {
-            rack_msg = msg_get(c_conn, msg->request, msg->data_store,
-                               __FUNCTION__);
+            rack_msg = msg_get(c_conn, msg->request, __FUNCTION__);
             if (rack_msg == NULL) {
                 log_debug(LOG_VERB, "whelp, looks like yer screwed "
                         "now, buddy. no inter-rack messages for "
@@ -790,7 +789,7 @@ req_forward_remote_dc(struct context *ctx, struct conn *c_conn, struct msg *msg,
     if (rack == NULL)
         rack = array_get(&dc->racks, 0);
 
-    struct msg *rack_msg = msg_get(c_conn, msg->request, msg->data_store, __FUNCTION__);
+    struct msg *rack_msg = msg_get(c_conn, msg->request, __FUNCTION__);
     if (rack_msg == NULL) {
         log_debug(LOG_VERB, "whelp, looks like yer screwed now, buddy. no inter-rack messages for you!");
         msg_put(rack_msg);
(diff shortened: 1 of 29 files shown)


src/dyn_conf.h
```diff
@@ -126,7 +71,7 @@ struct conf_pool {
     int                server_connections;    /* server_connections: */
     msec_t             server_retry_timeout_ms;  /* server_retry_timeout: in msec */
     int                server_failure_limit;  /* server_failure_limit: */
-    struct array       server;                /* servers: conf_server array */
+    struct conf_server *conf_datastore;
```
ipapapa: Add comment about the conf_datastore similar to the rest.

src/dyn_core.c
```diff
@@ -66,17 +63,18 @@ core_ctx_create(struct instance *nci)
 	/* initialize server pool from configuration */
 	status = server_pool_init(&ctx->pool, &ctx->cf->pool, ctx);
 	if (status != DN_OK) {
-		loga("Failed to initialize server pool!!!");
+		loga("Failed to initialize server pool!!! ... uninitializing");
```
ipapapa: We are unitializing every time status!=DN_OK in dyn_core. Let us keep the loga consistent.

src/dyn_server.c
```diff
@@ -114,65 +114,10 @@ server_active(struct conn *conn)
...
-	struct server_pool *sp = data;
-
-	s->owner = sp;
-
-	return DN_OK;
-}
-
 static void
-server_deinit(struct array *server)
+server_deinit(struct server **pdatastore)
```
ipapapa: so we do not use server_deinit anymore?

src/dyn_client.c
```diff
@@ -638,7 +638,7 @@ local_req_forward(struct context *ctx, struct conn *c_conn, struct msg *msg,
 
     conn_enqueue_inq(ctx, s_conn, msg);
     req_forward_stats(ctx, s_conn->owner, msg);
-    if(msg->data_store==DATA_REDIS){
+    if(g_data_store == DATA_REDIS){
```
ipapapa: why did we need this change?

src/dyn_server.c
shailesh33: I did the right thing (almost)

src/dyn_core.h
```diff
@@ -250,15 +236,14 @@ struct server {
...
-    uint32_t           idx;                  /* pool index */
     struct context     *ctx;                 /* owner context */
     struct conf_pool   *conf_pool;           /* back reference to conf_pool */
 
     struct conn        *p_conn;              /* proxy connection (listener) */
     uint32_t           dn_conn_q;            /* # client connection */
     struct conn_tqh    c_conn_q;             /* client connection q */
 
-    struct array       server;               /* server[] */
+    struct server      *datastore;               /* server[] */
```
ipapapa: If we change the name from server->datastore, then why we do not change the struct name as well?

src/dyn_client.c
shailesh33: we do not have multiple datastores. So no point in keeping that in every connection and message.

shailesh33: I am trying to make the connection and message more sleek

src/dyn_dnode_msg.c
```diff
@@ -31,7 +31,7 @@ dyn_parse_core(struct msg *r)
 {
    struct dmsg *dmsg;
    struct mbuf *b;
-   uint8_t *p, *token;
+   uint8_t *p = r->pos, *token;
```
ipapapa: what was the issue before?

src/dyn_core.h
shailesh33: that will be the next step yanni. Thanks for looking so closely. Today server and peer are the same structs. That should not be the case. but I guessed that change would be more exhaustive

src/dyn_dnode_msg.c
shailesh33: an uninitialized variable. thats it.

src/dyn_core.h
ipapapa: so are we leaving this change for the future?

PR #296, opened 2016-05-19
JSON last comma fix and removed unused variables

ipapapa:


src/dyn_stats.c
```diff
@@ -451,7 +449,7 @@ stats_add_string(struct stats_buffer *buf, struct string *key, struct string *va
 }
 
 static rstatus_t
-stats_add_num(struct stats_buffer *buf, struct string *key, int64_t val)
+stats_add_num_last(struct stats_buffer *buf, struct string *key, int64_t val, short last)
```
shailesh33: the param should be bool/boolean

src/dyn_stats.c
```diff
@@ -472,6 +476,18 @@ stats_add_num(struct stats_buffer *buf, struct string *key, int64_t val)
     return DN_OK;
 }
 
+
+static rstatus_t
+stats_add_num(struct stats_buffer *buf, struct string *key, int64_t val)
+{
+	if (stats_add_num_last(buf,key, val, 0) == DN_ERROR) {
```
shailesh33: use false, not 0

ipapapa: done

src/dyn_stats.c
ipapapa: done

PR #297, opened 2016-05-23
Peer struct

shailesh33:
Most importantly reduce the message size significantly, and seperate structs for datastore and peer node.


ipapapa:
Is this part of the re-org?

shailesh33:
Pretty much.

src/dyn_conf.c
```diff
@@ -148,20 +148,20 @@ conf_server_deinit(struct conf_server *cs)
 }
 
 // copy from struct conf_server to struct server
```
ipapapa: change comment to the new names

src/dyn_conf.c
```diff
@@ -386,19 +386,25 @@ conf_pool_transform(struct server_pool *sp, struct conf_pool *cp)
     sp->next_rebuild = 0ULL;
 
     sp->name = cp->name;
-    sp->addrstr = cp->listen.pname;
-    sp->port = (uint16_t)cp->listen.port;
+    sp->proxy_endpoint.pname = cp->listen.pname;
```
ipapapa: what is the difference between endpoint and proxy_endpoint?

src/dyn_connection.c
```diff
@@ -393,6 +404,185 @@ conn_deinit(void)
...
+    }
+
+    status = listen(p->sd, pool->backlog);
+    if (status < 0) {
+        log_error("listen on p %d on addr '%.*s' failed: %s", p->sd,
+                  p->pname.len, p->pname.data, strerror(errno));
+        return DN_ERROR;
+    }
+
+    status = dn_set_nonblocking(p->sd);
```
ipapapa: I never understood why in some cases we use "dn" and in some others "dyn"...

src/dyn_core.h
```diff
@@ -205,18 +205,33 @@ struct datacenter {
...
+    struct conn_tqh    s_conn_q;      /* server connection q */
+
+    msec_t             next_retry;    /* next retry time in msec */
+    uint32_t           failure_count; /* # consecutive failures */
+};
+
+struct node {
+    uint32_t           idx;           /* server index */
+    struct server_pool *owner;        /* owner pool */
+    struct endpoint     endpoint;
```
ipapapa: I am assuming you added "endpoint". Can you add some explanation why we wanted this? We are now pointing to a structure of structure...

src/dyn_dnode_peer.c
```diff
@@ -1519,7 +1427,7 @@ dnode_rsp_forward(struct context *ctx, struct conn *peer_conn, struct msg *rsp)
                   peer_conn->sd, rsp->dmsg->id, req->id, req->parent_id, rsp->id,
                   rsp->parent_id);
         if (c_conn && conn_to_ctx(c_conn))
-            stats_pool_incr(conn_to_ctx(c_conn), c_conn->owner,
+            stats_pool_incr(conn_to_ctx(c_conn),
                     peer_mismatch_requests);
 
         // TODO : should you be worried about message id getting wrapped around to 0?
```
ipapapa: ???

src/dyn_gossip.c
```diff
@@ -193,7 +193,7 @@ write_number(uint8_t *pos, uint64_t num, int *count)
    write_number(pos, num / 10, count);
    write_char(pos + (*count), '0' + (num % 10));
    *count = *count + 1;
-   return count;
+   return *count;
```
ipapapa: Is gossip_msg_to_core supposed to return rstatus_t, i.e. DN_OK or DN_ERROR?

src/dyn_server.c
```diff
@@ -115,48 +116,78 @@ server_active(struct conn *conn)
...
+server_pool_run(struct server_pool *pool)
+{
+	ASSERT(pool->datastore != NULL);
+
+	switch (pool->dist_type) {
+	case DIST_KETAMA:
+		return ketama_update(pool);
+
+	case DIST_VNODE:
+		//return vnode_update(pool);
```
ipapapa: Shall we comment out the whole case DIST_VNODE and let the switch reach NOT_REACHED()?

src/dyn_conf.c
shailesh33: will do that.

src/dyn_conf.c
shailesh33: endpoint is the type. I use the same name for variable unless there are multiple which in ourcase is proxy for 8102 and dyn_proxy which is 8101

src/dyn_connection.c
shailesh33: hmm thats a puzzle for me too. But this is what I think, the original code has nc_ and so they thought we also want a 2 letter prefix so it became dn_ but then later dyn make to make it more explicit. Generally library calls are wrapped with dn_ and objects inside dynomite have dyn_. thats what I have observed.

src/dyn_conf.c
shailesh33: I am trying to make as less intrusive changes as possible

src/dyn_core.h
shailesh33: if you see everything in endpoint struct is repeated everywhere almost so it makes sense to depict a service as a "endpoint" to which we either listen to or connect to.

src/dyn_dnode_peer.c
shailesh33: stats_pool_incr and others dont take the server pool anymore.

src/dyn_server.c
shailesh33: I dont know. The funny part is all other cases of update like ketama updater, murmur update are all commented internally.

src/dyn_gossip.c
shailesh33: This is a different function... write_number ... returns int

src/dyn_conf.c
shailesh33: I did this in my latest code. Forgive me for not updating this PR.

src/dyn_server.c
ipapapa: Can you comment them out here the ones not used? Maybe add a comment that partial implementation is there or whatever. It would be nice to have the same way about this.

ipapapa:
The branch is out of date.

src/dyn_server.c
shailesh33: Done.

PR #300, opened 2016-06-01
Fix security issue

shailesh33:
We are using string functions which do the right thing during dup. So
lets use it. Otherwise we were writing an extra \0 sometimes causing
heap overflow

### DIFF
--- a/src/dyn_string.c
+++ b/src/dyn_string.c
@@ -75,14 +75,12 @@ string_duplicate(struct string *dst, const struct string *src)
     ASSERT(dst->len == 0 && dst->data == NULL);
     ASSERT(src->len != 0 && src->data != NULL);
 
-    dst->data = dn_strndup(src->data, src->len + 1);
+    dst->data = dn_strndup(src->data, src->len);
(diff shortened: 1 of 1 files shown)


src/dyn_string.c
```diff
@@ -92,13 +90,12 @@ string_copy(struct string *dst, const uint8_t *src, uint32_t srclen)
...
 
-    dst->data = dn_strndup(src, srclen + 1);
+    dst->data = dn_strndup(src, srclen);
     if (dst->data == NULL) {
         return DN_ENOMEM;
     }
 
-    dst->len = srclen;
-    dst->data[dst->len] = '\0';
+    dst->len = dn_strlen(dst->data);
```
timiblossom: dn_strlen uses strlen() to calculate the length of a string.  strlen() bases on the terminating null character to determine the length of a string.  So if data is not a null terminated string, how could dn_strlen(data) behave?

shailesh33: Strdup guarantees that it's a null terminated.

PR #301, opened 2016-06-13
anti-entropy core code

ipapapa:
This is the second part of the anti-entropy code. The main utility of the anti-entropy
- Anti-entropy encryption is based on key and iv provided in .pem files.
- epoll, kqueue, evport event loops for anti-entropy


src/dyn_conf.h
```diff
@@ -89,6 +89,8 @@ struct conf_pool {
     struct string      read_consistency;
     struct string      write_consistency;
     struct string      pem_key_file;
+    struct string      recon_key_file;       /* file with Key encryption in reconciliation */
+	struct string      recon_iv_file;        /* file with Initialization Vector encryption in reconciliation */
```
timiblossom: extra space?

src/dyn_conf.c
```diff
@@ -440,6 +446,8 @@ conf_pool_transform(struct server_pool *sp, struct conf_pool *cp)
 
     sp->secure_server_option = get_secure_server_option(cp->secure_server_option);
     sp->pem_key_file = cp->pem_key_file;
+    sp->recon_key_file = cp->recon_key_file;
```
shailesh33: use string_copy, I know we dont do that everywhere, but lets make sure the new code does the right thing

src/dynomite.c
```diff
@@ -46,6 +46,9 @@
 #define DN_STATS_ADDR       STATS_ADDR
 #define DN_STATS_INTERVAL   STATS_INTERVAL
 
+#define DN_ENTROPY_PORT		ENTROPY_PORT
```
shailesh33: Can you point me to where ENTROPY_PORT is defined?

src/event/dyn_epoll.c
```diff
@@ -350,4 +350,52 @@ event_loop_stats(event_stats_cb_t cb, void *arg)
...
 
+void
+event_loop_entropy(event_entropy_cb_t cb, void *arg)
+{
+    struct entropy *st = arg;
+    int status, ep;
+    struct epoll_event ev;
+    st->interval = 30;
+
+    ep = epoll_create(1);
```
shailesh33: what is the 1 parameter?

src/event/dyn_epoll.c
```diff
@@ -350,4 +350,52 @@ event_loop_stats(event_stats_cb_t cb, void *arg)
     ep = -1;
 }
 
+void
+event_loop_entropy(event_entropy_cb_t cb, void *arg)
+{
+    struct entropy *st = arg;
```
shailesh33: can you rename the variable from st to something related to entropy?

src/dynomite.c
ipapapa: src/entropy/dyn_entropy.h

src/event/dyn_epoll.c
```diff
@@ -350,4 +350,52 @@ event_loop_stats(event_stats_cb_t cb, void *arg)
...
+
+        cb(st, &n);
+    }
+
+error:
+    status = close(ep);
+    if (status < 0) {
+        log_error("close e %d failed, ignored: %s", ep, strerror(errno));
+    }
+    ep = -1;
```
shailesh33: what happens to Dynomite if this fails, reconciliation will never work?

src/dyn_core.c
```diff
@@ -126,18 +126,33 @@ core_event_base_create(struct context *ctx)
...
 static rstatus_t
+core_entropy_init(struct context *ctx)
+{
+    struct instance *nci = ctx->instance;
+    /* initializing anti-entropy */
+    ctx->entropy = entropy_init(ctx, nci->entropy_port, nci->entropy_addr);
+    if (ctx->entropy == NULL) {
+    	log_error("Failed to create entropy!!!");
+    }
+	rstatus_t status = core_event_base_create(ctx);
```
timiblossom: Indentation

src/dyn_core.c
```diff
@@ -126,18 +126,33 @@ core_event_base_create(struct context *ctx)
...
     struct instance *nci = ctx->instance;
 	ctx->stats = stats_create(nci->stats_port, nci->stats_addr, nci->stats_interval,
 			                  nci->hostname, &ctx->pool, ctx);
     if (ctx->stats == NULL) {
-		loga("Failed to create stats!!!");
+    	log_error("Failed to create stats!!!");
 		return DN_ERROR;
 	}
-    rstatus_t status = core_event_base_create(ctx);
+    rstatus_t status = core_entropy_init(ctx);
```
timiblossom: We should run the anti-entropy tasks in a separate thread, not this main one.

ipapapa: anti-entropy tasks are run on a separate thread. This PR is about entropy initialization only.

ipapapa: See here for more information about thread creation and callbacks: https://github.com/Netflix/dynomite/blob/dev/src/entropy/dyn_entropy_util.c#L652-L675

src/event/dyn_epoll.c
ipapapa: correct, unless Dynomite is restarted.

shailesh33: then return an error so dynomite does not start?

src/dyn_conf.h
ipapapa: Interesting... the IDE hides the indentation, but it was there in vi. I will fix it.

src/dyn_conf.c
ipapapa: We use string_copy_c when we copy to cp->recon_key_file. There was a warning there that I fixed.

src/event/dyn_epoll.c
ipapapa: Renamed `st` to `ent`.

src/dyn_core.c
ipapapa: fixed it. dyno_core.c seems to be overall really bad with indentation.

src/event/dyn_epoll.c
ipapapa: Dynomite will not start. Looking into the chain of calls: `event_loop_entropy(event_entropy_cb_t cb, void *arg)` is used inside `void *
entropy_loop(void *arg)` which is correspondingly used in `status = pthread_create(&cn->tid, NULL, entropy_loop, cn);`. The pt_thread API third argument is void *(*start_routine) (void *) http://man7.org/linux/man-pages/man3/pthread_create.3.html. It seems that even if we return a DN_ERROR, it would not be useful. We followed here the same approach as event_loop_stats.

PR #302, opened 2016-06-29
Implement Safe Quorum

shailesh33:
o Implement safe quorum which essentially means that we fail a
request if the quorum responses do not match in checksums
    o Store checksums in response manager and compute them as we receive
responses, this is avoid duplicate computation that we had
    o the new option is DC_SAFE_QUORUM instead of DC_QUORUM_SAFE so that string comparison is
less error prone
    o Add new conf file to create a 3 rack cluster in test
    o Implement dn_strerror to get error string for dynomite specific errors
    o Added command options in dyn_stats

### DIFF
--- a/conf/redis_rack1_node.yml
+++ b/conf/redis_rack1_node.yml
@@ -12,5 +12,5 @@ dyn_o_mite:
   secure_server_option: datacenter
   pem_key_file: conf/dynomite.pem
   data_store: 0
-  read_consistency : DC_SAFE_QUORUM
-  write_consistency : DC_SAFE_QUORUM
+  read_consistency : DC_ONE
+  write_consistency : DC_ONE
--- a/src/dyn_dnode_peer.c
+++ b/src/dyn_dnode_peer.c
@@ -462,7 +462,7 @@ dnode_peer_ack_err(struct context *ctx, struct conn *conn, struct msg *req)
 {
     if ((req->swallow && req->noreply) || // no reply
         (req->swallow && (req->consistency == DC_ONE)) || // dc one
-        (req->swallow && (req->consistency == DC_QUORUM) // remote dc request
+        (req->swallow && ((req->consistency == DC_QUORUM) || (req->consistency == DC_SAFE_QUORUM)) // remote dc request
                       && (!conn->same_dc)) ||
         (req->owner == conn)) // a gossip message that originated on this conn
     {
@@ -1462,12 +1462,12 @@ dnode_rsp_forward_match(struct context *ctx, struct conn *peer_conn, struct msg 
 
     /* if client consistency is dc_quorum, forward the response from only the
        local region/DC. */
-    if ((req->consistency == DC_QUORUM) && !peer_conn->same_dc) {
+    if (((req->consistency == DC_QUORUM) || (req->consistency == DC_SAFE_QUORUM))
+        && !peer_conn->same_dc) {
         if (req->swallow) {
             dnode_rsp_swallow(ctx, peer_conn, req, rsp);
             return;
         }
-        //log_warn("req %d:%d with DC_QUORUM consistency is not being swallowed");
     }
 
     log_debug(LOG_DEBUG, "DNODE RSP RECEIVED %s %d dmsg->id %u req %u:%u rsp %u:%u, ",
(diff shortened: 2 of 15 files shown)


src/dyn_stats.c
```diff
@@ -1045,18 +1045,22 @@ parse_request(int sd, struct stats_cmd *st_cmd)
...
                     } else if (strncmp(op, "/write", 6) == 0) {
                         char* type = op + 6;
-                        if (strcmp(type, "/dc_one") == 0)
+                        if (!dn_strcasecmp(type, "/"CONF_STR_DC_ONE))
                             g_write_consistency = DC_ONE;
-                        else if (strcmp(type, "/dc_quorum") == 0)
+                        else if (!dn_strcasecmp(type, "/"CONF_STR_DC_QUORUM))
                             g_write_consistency = DC_QUORUM;
+                        else if (!dn_strcasecmp(type, "/"CONF_STR_DC_SAFE_QUORUM))
+                            g_write_consistency = DC_SAFE_QUORUM;
```
ipapapa: Why DC_SAFE_QUORUM and not DC_QUORUM_SAFE?

ipapapa:
Can you also update the Wiki for new consistency properties? Are we planning to support DC_ALL or not?

conf/redis_rack1_node.yml
```diff
@@ -4,10 +4,13 @@ dyn_o_mite:
...
+  - 127.0.0.3:8101:rack3:dc:1383429731
   listen: 127.0.0.1:8102
   servers:
-  - 127.0.0.1:22123:1
+  - 127.0.0.1:22121:1
   tokens: '1383429731'
   secure_server_option: datacenter
   pem_key_file: conf/dynomite.pem
   data_store: 0
+  read_consistency : DC_SAFE_QUORUM
```
ipapapa: Is there a reason why we default to DC_SAFE_QUORUM and not DC_ONE?

src/dyn_dnode_peer.c
```diff
@@ -1462,12 +1462,12 @@ dnode_rsp_forward_match(struct context *ctx, struct conn *peer_conn, struct msg
 
     /* if client consistency is dc_quorum, forward the response from only the
```
ipapapa: comment must include dc_safe_quorum as well.

src/dyn_response_mgr.c
```diff
@@ -103,42 +104,51 @@ rspmgr_get_response(struct response_mgr *rspmgr)
...
     uint32_t chk0, chk1, chk2;
-    chk0 = msg_payload_crc32(rspmgr->responses[0]);
-    chk1 = msg_payload_crc32(rspmgr->responses[1]);
+    chk0 = rspmgr->checksums[0];
+    chk1 = rspmgr->checksums[1];
     if (chk0 == chk1) {
         return rspmgr->responses[0];
-    } else {
-        chk2 = msg_payload_crc32(rspmgr->responses[2]);
+    } else if (rspmgr->good_responses == 3) {
```
ipapapa: In order to achieve quorum (safe or not safe), we need at least 2 good responses. Why are waiting for 3 good responses?

src/dyn_message.h
```diff
@@ -205,15 +205,28 @@ typedef enum dyn_error {
...
+    NO_QUORUM_ACHIEVED,
 } dyn_error_t;
 
+static inline char *
+dn_strerror(dyn_error_t err)
+{
+    switch(err)
+    {
+        case NO_QUORUM_ACHIEVED:
+            return "Failed to achieve Quorum";
```
ipapapa: I was wondering whether we want to add some descriptive in the error message that this is generated by Dynomite. Like "DN" or similar. What do you think?

src/dyn_stats.c
shailesh33: the dn_strcasecmp stupidly does a strncasecmp and so we have to be careful to compare quorum_safe before quorum. SO I just renamed it

conf/redis_rack1_node.yml
shailesh33: hmm. good question. Is anyone using these files? If so I can revert it to DC_ONE

src/dyn_response_mgr.c
shailesh33: we wait for 3rd only after first 2 checksum do not match

src/dyn_message.h
shailesh33: The source field we see is "Peer". I think that can be changed

conf/redis_rack1_node.yml
shailesh33: Done.

src/dyn_dnode_peer.c
shailesh33: done

src/dyn_message.h
shailesh33: Done.

sbirari@ubuntu:~/Work/dynomite_temp/src (hotfix/safequorum)*$ redis-cli -h 127.0.0.1 -p 8102
127.0.0.1:8102> get shailesh
(error) ERR Dynomite: Failed to achieve Quorum
127.0.0.1:8102>

src/dyn_response_mgr.c
```diff
@@ -103,42 +104,51 @@ rspmgr_get_response(struct response_mgr *rspmgr)
...
         msg_dump(rspmgr->msg);
     }
     if (log_loggable(LOG_VVERB)) {
         log_error("Respone 0: ");
         msg_dump(rspmgr->responses[0]);
         log_error("Respone 1: ");
         msg_dump(rspmgr->responses[1]);
-        log_error("Respone 2: ");
-        msg_dump(rspmgr->responses[2]);
+        if (rspmgr->good_responses == 3) {
```
timiblossom: I seems the algorithm always expects the deployments are 3 replications.  Should we change this or document it?

ipapapa: It would also be nice not to start Dynomite, if the user sets RF!=3 and DC_QUORUM_SAFE.

PR #313, opened 2016-08-04
Fix Duplicate dnode headers while sending response to peers

shailesh33:
o Add a field in message to remember if we prepended a dnode header
while sending a response to peer
    o Check for this field before prepending the header

### DIFF
--- a/src/dyn_message.h
+++ b/src/dyn_message.h
@@ -323,6 +323,12 @@ struct msg {
     unsigned             first_fragment:1;/* first fragment? */
     unsigned             last_fragment:1; /* last fragment? */
     unsigned             swallow:1;       /* swallow response? */
+    /* A hack here: We need a way in dnode_rsp_send_next to remmeber if we already
+     * did a dmsg_write of a dnode header in this message. If we do not remember it,
+     * then if the same message gets attempted to be sent twice in msg_send_chain,
+     * (due to lack of space in the previous attempt), we will prepend another header
+     * and we will have corrupted message at the destination */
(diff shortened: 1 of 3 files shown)


src/dyn_message.h
```diff
@@ -323,6 +323,12 @@ struct msg {
     unsigned             first_fragment:1;/* first fragment? */
     unsigned             last_fragment:1; /* last fragment? */
     unsigned             swallow:1;       /* swallow response? */
+    /* A hack here: We need a way in dnode_rsp_send_next to remmeber if we already
```
ipapapa: type in remmeber

src/dyn_message.h
```diff
@@ -323,6 +323,12 @@ struct msg {
...
+     * did a dmsg_write of a dnode header in this message. If we do not remember it,
+     * then if the same message gets attempted to be sent twice in msg_send_chain,
+     * (due to lack of space in the previous attempt), we will prepend another header
+     * and we will have corrupted message at the destination */
+    unsigned             dnode_header_prepended:1;       /* swallow response? */
```
ipapapa: Are you taking the description from `swallow:1`?

ipapapa:
other than the minor comments, LGTM.

timiblossom:
Why do we label it as a bug? I thought it is an enhancement?  Other than that, looks good to me.

ipapapa:
It was a bug seen in production when performing `LRANGE my_list 0 -1` in a `DC_QUORUM` consistency-level cluster.

PR #314, opened 2016-08-07
New latency metrics

shailesh33:
* New histograms to track time a message spends in queue of server,
local peer and remote peer
* Another set of histograms to track when a response is processed
for a given request. Note: this is not a network delay but the overall
processing delay

### DIFF
--- a/src/dyn_dnode_peer.c
+++ b/src/dyn_dnode_peer.c
@@ -720,7 +720,7 @@ dnode_peer_close_socket(struct context *ctx, struct conn *conn)
         log_debug(LOG_VERB, "In dnode_peer_close_socket");
     }
 
-    if ((conn != NULL) && (conn->sd != -1)) {
+    if (conn != NULL) {
         status = close(conn->sd);
         if (status < 0) {
             log_error("dyn: close s %d failed, ignored: %s", conn->sd, strerror(errno));
--- a/src/dyn_dnode_peer.c
+++ b/src/dyn_dnode_peer.c
@@ -716,11 +716,11 @@ static void
 dnode_peer_close_socket(struct context *ctx, struct conn *conn)
 {
     rstatus_t status;
-        if (log_loggable(LOG_VERB)) {
-       log_debug(LOG_VERB, "In dnode_peer_close_socket");
-        }
+    if (log_loggable(LOG_VERB)) {
+        log_debug(LOG_VERB, "In dnode_peer_close_socket");
+    }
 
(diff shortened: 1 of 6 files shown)


src/dyn_message.h
```diff
@@ -272,7 +272,8 @@ struct msg {
     struct msg           *peer;           /* message peer */
     struct conn          *owner;          /* message owner - client | server */
     int64_t              stime_in_microsec;  /* start time in microsec */
-    int64_t              remote_region_send_time; /* time in microsec when message sent to remote region */
+    int64_t              request_inqueue_enqueue_time_us; /* when message was enqueued in inqueue, either to a redis server or remote region or cross rack */
+    int64_t              request_send_time; /* when message was sent: either to a redis server or remote region or cross rack */
```
ipapapa: change the comment of "redis server" to something generic.

src/dyn_stats.c
```diff
@@ -458,9 +458,10 @@ stats_create_bufs(struct stats *st)
 
     /* footer */
     size += 2;
+    // Accomodate for new fields that are directly added using stats_add_num_str
+    size += 1024;
```
ipapapa: I am not following why we add 1024 here?

ipapapa:
We need to consider merging our hotfixes to dev because we are diverging and it will be a difficult job later.

src/dyn_dnode_peer.c
```diff
@@ -716,11 +716,11 @@ static void
...
     rstatus_t status;
-        if (log_loggable(LOG_VERB)) {
-       log_debug(LOG_VERB, "In dnode_peer_close_socket");
-        }
+    if (log_loggable(LOG_VERB)) {
+        log_debug(LOG_VERB, "In dnode_peer_close_socket");
+    }
 
-    if (conn != NULL) {
+    if ((conn != NULL) && (conn->sd != -1)) {
```
ipapapa: A socket descriptor can be  -1 if an error has occurred in either `open()` or `creat()`. With the new addition in the if statement, it seems that we are not going to close the socket description if an error appeared on the `creat()`.

src/dyn_stats.c
shailesh33: We are not creating "string" object for the new metrics. Instead I am just using string literals. This is a rough guesstimate for the size of the new strings + size of the values. If you see the whole function is about guessing the total size of the stats output in bytes. And theres no reason for such detailed calculation. I am just taking a rough estimate and using it.

PR #316, opened 2016-08-09
Patch dev with fixes from Master

shailesh33:
Since dev and master diverged recently, manually fix the following
patches to dev
    o Implement Safe Quorum
    o Fix for dnode duplicate header problem while replicating huge
payloads
    o Addtional latencies for histograms

### DIFF
--- a/src/dyn_stats.h
+++ b/src/dyn_stats.h
@@ -252,8 +252,8 @@ struct stats {
     volatile int              aggregate;      /* shadow (b) aggregate? */
     volatile int              updated;        /* current (a) updated? */
     volatile bool             reset_histogram;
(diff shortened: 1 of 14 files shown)


src/dyn_stats.h
```diff
@@ -254,7 +254,15 @@ struct stats {
     volatile bool             reset_histogram;
     struct histogram latency_histo;
```
ipapapa: are these histograms also accessed by the two threads? If yes, then shall `latency_histo` and `payalod_size_histo` be volatile as well?

ipapapa:
Shouldn't we do the same for `volatile` histograms in the master branch?

PR #317, opened 2016-08-16
Fix latency spike during connection failures

shailesh33:
Fix latency spike during connection failures
    The issue was when a connection is not there, because a peer node is
down, we were queuing up the messages till they would timeout and fail
eventually. This would lead to higher overall average latency and much
higher 99th latencies. The solution is to not queue up messages and fail
the requests till there is a good connection
    o Drop the messages and respond with error immediately when a
connection is not available
    o Do the same when the connection is in a connecting state
    o Split the dnode_peer_conn into 2 functions to get the peer server
and then to get a connection to it
    o Mark the node as down after some failure attempts. Implement exponential backoff upto 10 secs to reconnect to peer

### DIFF
--- a/src/dyn_client.c
+++ b/src/dyn_client.c
@@ -656,57 +656,84 @@ static void
 admin_local_req_forward(struct context *ctx, struct conn *c_conn, struct msg *msg,
                         struct rack *rack, uint8_t *key, uint32_t keylen)
 {
-    struct conn *p_conn;
-
     ASSERT((c_conn->type == CONN_CLIENT) ||
            (c_conn->type == CONN_DNODE_PEER_CLIENT));
 
-    p_conn = dnode_peer_pool_conn(ctx, c_conn->owner, rack, key, keylen, msg->msg_type);
+    struct server *peer = dnode_peer_pool_server(ctx, c_conn->owner, rack, key, keylen, msg->msg_type);
+    if (!peer->is_local) {
+        send_rsp_integer(ctx, c_conn, msg);
+        return;
+    }
+
+    struct conn *p_conn = dnode_peer_pool_server_conn(ctx, peer);
     if (p_conn == NULL) {
         c_conn->err = EHOSTDOWN;
         req_forward_error(ctx, c_conn, msg, c_conn->err);
         return;
     }
 
-    struct server *peer = p_conn->owner;
-
-    if (peer->is_local) {
-        send_rsp_integer(ctx, c_conn, msg);
-    } else {
-        log_debug(LOG_NOTICE, "Need to delete [%.*s] ", keylen, key);
-        local_req_forward(ctx, c_conn, msg, key, keylen);
-    }
+    log_debug(LOG_NOTICE, "Need to delete [%.*s] ", keylen, key);
+    local_req_forward(ctx, c_conn, msg, key, keylen);
 }
 
 void
 remote_req_forward(struct context *ctx, struct conn *c_conn, struct msg *msg, 
-                        struct rack *rack, uint8_t *key, uint32_t keylen)
+                   struct rack *rack, uint8_t *key, uint32_t keylen)
 {
-    struct conn *p_conn;
-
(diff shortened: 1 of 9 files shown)


src/dyn_client.c
```diff
@@ -656,57 +656,85 @@ static void
...
+        }
+        // No response for remote dc
+        struct server_pool *pool = c_conn->owner;
+        bool same_dc = is_same_dc(pool, peer)? 1 : 0;
+        if (!same_dc) {
+            msg_put(msg);
+            return;
+        }
+        // All other cases return a response
+        //log_warn("Directly sending a response");
```
ipapapa: if we are not using the warning message, is there a reason to keep it around?

src/dyn_dnode_peer.c
```diff
@@ -513,33 +515,39 @@ dnode_peer_failure(struct context *ctx, struct server *server)
...
+             pool->name.data, server->pname.len, server->pname.data,
+             WAIT_BEFORE_UPDATE_PEERS_IN_MILLIS / 1000 );
 
-    //if (server->failure_count == 3)
-    //   server->next_retry = now + WAIT_BEFORE_RECONNECT_IN_MILLIS;
+    stats_pool_incr(ctx, pool, peer_ejects);
 
-    status = dnode_peer_pool_run(pool);
+    //server->failure_count = 0;
+    server->next_retry = now + (server->reconnect_backoff_s * 1000 * 1000);
```
ipapapa: sounds weird all these changes of units. Can we have `next_retry_ms`?

src/dyn_dnode_peer.c
```diff
@@ -1246,16 +1253,38 @@ dnode_peer_pool_conn(struct context *ctx, struct server_pool *pool,
...
 
+struct conn *
+dnode_peer_pool_server_conn(struct context *ctx, struct server *server)
+{
+    // This peer server is marked as down. lets check if it is time to connect
+    if (server->state == DOWN) {
+        int64_t now = dn_usec_now();
+        static int64_t next_log = 0; // Log every 1 sec
+        if (server->next_retry && (now > server->next_retry)) {
+            server->state = NORMAL; // Reset the connection status if any
```
ipapapa: your comment and the `server->state` disagree. Are you `REST` or making it `NORMAL`?

src/dyn_dnode_peer.h
```diff
@@ -9,18 +9,20 @@
 #ifndef _DYN_DNODE_PEER_H_
 #define _DYN_DNODE_PEER_H_
 
-#define WAIT_BEFORE_RECONNECT_IN_MILLIS      30000
+#define MAX_WAIT_BEFORE_RECONNECT_IN_SECS    10
```
ipapapa: so again, one unit is seconds and the other milliseconds below.

src/dyn_message.c
```diff
@@ -256,9 +256,9 @@ _msg_get(struct conn *conn, const char *const caller)
 
     alloc_msg_count++;
 
-
-    log_warn("alloc_msg_count: %lu caller: %s conn: %s sd: %d",
-             alloc_msg_count, caller, conn_get_type_string(conn), conn->sd);
+    if (alloc_msg_count % 1000 == 0)
```
ipapapa: where did the 1000 come from? Can we `#define` in the dyn_message.c with a representative name?

ipapapa:
tests are failing :)

src/dyn_message.c
shailesh33: I forgot to take care of this. But it is to reduce the alloc msg count logging that we see.

src/dyn_dnode_peer.h
timiblossom: Is it configurable? 10 seconds seems too short.  Basically with 2 as the multiplier factor, we can only retry 3 times max.

shailesh33: that is just the max wait between retries. We keep on trying every 10 sec after that

PR #326, opened 2016-08-26
An important patch for a few things

shailesh33:
o Set peer pointers appropriately at multiple places
    o Call conn_handle_response instead of client_handle_response
directly
    o Fix Replication for 2 Hop case when we reroute replication to
remote DC to a different node

### DIFF
--- a/src/dyn_crypto.c
+++ b/src/dyn_crypto.c
@@ -406,7 +406,7 @@ dyn_aes_encrypt_msg(struct msg *msg, unsigned char *aes_key)
             return DN_ERROR;
         }
 
-        int n = dyn_aes_encrypt(mbuf->start, mbuf->last - mbuf->start, nbuf, aes_key);
+        int n = dyn_aes_encrypt(mbuf->pos, mbuf->last - mbuf->pos, nbuf, aes_key);
         if (n > 0)
             count += n;
 
--- a/src/dyn_dnode_client.c
+++ b/src/dyn_dnode_client.c
@@ -374,6 +374,52 @@ dnode_rsp_send_next(struct context *ctx, struct conn *conn)
 {
     rstatus_t status;
 
+    // SMB: There is some non trivial thing happening here. And I think it is very
+    // important to read this before anything is changed in here. There is also a
+    // bug that exists which I will mention briefly:
+    // A message is a structure that has a list of mbufs which hold the actual data.
+    // Each mbuf has start, pos, last as pointers (amongst others) which indicate start of the
+    // buffer, current read position and end of the buffer respectively.
+    //
+    // Every time a message is sent to a peer within dynomite, a DNODE header is
+    // prepended which is created using dmsg_write. A message remembers this case
+    // in dnode_header_prepended, so that if the messsage is sent in parts, the
+    // header is not prepended again for the subsequent parts.
+    //
+    // Like I said earlier there is a pos pointer in mbuf. If a message is sent
+    // partially (or it is parsed partially too I think) the pos reflects that
+    // case such that things can be resumed where it left off.
+    //
+    // dmsg_write has a parameter which reflects the payload length following the
+    // dnode header calculated by msg_length. msg_length is a summation of all
+    // mbuf sizes (last - start). Which I think is wrong.
+    //
+    // +------------+           +---------------+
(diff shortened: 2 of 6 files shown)


src/dyn_dnode_client.c
```diff
@@ -374,6 +374,52 @@ dnode_rsp_send_next(struct context *ctx, struct conn *conn)
...
+    // this header and prepend its own header for sending it back to DC1N1. This
+    // gets handled in encryption case since we overwrite all mbufs in the response
+    // However if the encryption is off, the message length sent to dmsg_write
+    // consists of the header from DC2N2 also which is wrong. So this relaying
+    // of responses will not work for the case where encryption is disabled.
+    //
+    // So msg_length should really be from mbuf->pos and not mbuf->start. This
+    // is a problem only with remote region replication since that is the only
+    // case where we CAN have 2 hops to send the request/response. This is also
+    // not a problem if encryption is ON.
```
ipapapa: Are we effectively saying that encryption OFF may create issues? If this is the case, then we need to disable the  property and have it "enabled" by default, or change the implementation to handle this case.

src/dyn_dnode_client.c
```diff
@@ -374,6 +374,52 @@ dnode_rsp_send_next(struct context *ctx, struct conn *conn)
...
+    // in dnode_header_prepended, so that if the messsage is sent in parts, the
+    // header is not prepended again for the subsequent parts.
+    //
+    // Like I said earlier there is a pos pointer in mbuf. If a message is sent
+    // partially (or it is parsed partially too I think) the pos reflects that
+    // case such that things can be resumed where it left off.
+    //
+    // dmsg_write has a parameter which reflects the payload length following the
+    // dnode header calculated by msg_length. msg_length is a summation of all
+    // mbuf sizes (last - start). Which I think is wrong.
```
ipapapa: is it `last-start` or `last-pos`?

shailesh33: it is last-start currently which I intend to change to last-pos. With that it will fix the case with no encryption

shailesh33: I will have to test the encryption off case, but I think that is a bug.

src/dyn_dnode_client.c
shailesh33: thats my gutt feeling. I am going to test it out. If it works I will move this comment

src/dyn_crypto.c
```diff
@@ -406,7 +406,7 @@ dyn_aes_encrypt_msg(struct msg *msg, unsigned char *aes_key)
             return DN_ERROR;
         }
 
-        int n = dyn_aes_encrypt(mbuf->start, mbuf->last - mbuf->start, nbuf, aes_key);
+        int n = dyn_aes_encrypt(mbuf->pos, mbuf->last - mbuf->pos, nbuf, aes_key);
```
timiblossom: Does it make any difference here because mbuf->start is always the same as pos in this case?

src/dyn_dnode_client.c
timiblossom: This would be a good catch.  However, I recall originally that there is no need for such response from DC2N2 -> DC2N1 -> DC1N1 as we only need DC2N1 -> DC1N1.
Whatever happen within the same region, it is the responsibility of the coordinator (DC2N1 in this case) to make sure data get replicated.

I don't mind which direction we should take but we need to document the behavior.  

I wonder what the memory foot print would look like when the request-response in-memory queue gets piled up due to 1 extra hop delay of the responses.

shailesh33: You have a point. But the thing is, if DC2N1 is not the token owner, the request gets forwarded to DC2N2 and whose response is relayed back. So the response between DCs is not a plain ack saying "Hey I got your message, dont worry, I will take it from here". But it is a valid response saying "Hey the write you requested, succeeded on one node". Thats the way it has been always.

src/dyn_crypto.c
shailesh33: It makes a difference. Start and pos are same for normal cases but for a dnode response, start points to the start of the dnode header and pos would point to the payload because by now the dnode header would have already parsed.

shailesh33: All other cases, start and pos are the same

PR #347, opened 2016-09-29
stats test

danielhochman:


shailesh33:
Hmm... looks like the output is not a json format? Did this work for you?

danielhochman:
that's the point of the test. this is broken on dev and master.

```
"redis_req_other":366,
}}}
```

shailesh33:
I think this was fixed. Let me check it got lost

travis.sh
```diff
@@ -84,6 +84,13 @@ sleep 10
 RESULT=$?
 echo $RESULT
 
+# check a single stats port
+curl -s localhost:22222 | python -mjson.tool > /dev/null
```
shailesh33: it should be curl -s http://localhost:22222/info

ipapapa:
I tried on `dev` branch and as @shailesh33 mentioned test passes with `info`.

travis.sh
danielhochman: fixed

mikramulhaq:
@shailesh33 @ipapapa @danielhochman 
#348 regardless needs to go in. Below code is fundamentally wrong, what is pos[-1] here?

// eliminate the , if any
    if (pos[-1] == ',') {
        buf->len--;
        pos--;
    }

shailesh33:
Yes you are probably right. Why do we get white space at the end?

mikramulhaq:
It was hard to follow through the code, so I did not spend too much time investigating.

But if we are removing last comma, no harm in checking if there are any white spaces.

PR #350, opened 2016-09-30
Fixed the token generated python script

balajivenki:
Fixed the script to generate tokens equidistant to each other and max as 4294967295

### DIFF
--- a/scripts/dynomite/generate_yamls.py
+++ b/scripts/dynomite/generate_yamls.py
@@ -14,19 +14,21 @@ APPNAME='dyn_o_mite'
 CLIENT_PORT='8102'
 DYN_PEER_PORT=8101
 MEMCACHE_PORT='11211'
+MAX_TOKEN = 4294967295
 
 DEFAULT_DC = 'default_dc'
 
 # gen map of node to random count (3-7) of random tokens (0-MAX_INT)
 token_map = dict()
+token_item = (MAX_TOKEN // (len(sys.argv) -1))
 for i in range(1, len(sys.argv)):
     node = sys.argv[i]
-    token_cnt = 1 #random.randrange(2,7, 1)
     tokens = []
-    for j in range(token_cnt):
-        t = random.randint(0,4294967295)
-        tokens.append(t)
-    tokens.sort()
+    token_value = (token_item * i)
+    if token_value > MAX_TOKEN:
+        token_value = MAX_TOKEN
+
+    tokens.append(token_value)
 
     tok_str = ','.join(str(it) for it in tokens)
     token_map[node] = tok_str
@@ -59,5 +61,3 @@ for k,v in token_map.items():
     file_name = ip_dc[0] + '.yml'
     with open(file_name, 'w') as outfile:
         outfile.write( yaml.dump(outer, default_flow_style=False) )
-
-    
--- a/scripts/dynomite/generate_yamls.py
+++ b/scripts/dynomite/generate_yamls.py
@@ -2,13 +2,13 @@
 
 '''
 script for generating dynomite yaml files for every node in a cluster.
-tokens randomly generated, with a variable count per node, as well.
-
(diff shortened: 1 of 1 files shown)


scripts/dynomite/generate_yamls.py
```diff
@@ -14,19 +14,21 @@
 CLIENT_PORT='8102'
 DYN_PEER_PORT=8101
 MEMCACHE_PORT='11211'
+MAX_TOKEN = 4294967295
 
 DEFAULT_DC = 'default_dc'
 
 # gen map of node to random count (3-7) of random tokens (0-MAX_INT)
```
timiblossom: Should remove this comment.

scripts/dynomite/generate_yamls.py
```diff
@@ -14,19 +14,21 @@
 CLIENT_PORT='8102'
```
timiblossom: Maybe add an example how to run it and a short description

balajivenki: Its there at the top @timiblossom

scripts/dynomite/generate_yamls.py
balajivenki: Done

scripts/dynomite/generate_yamls.py
```diff
@@ -14,19 +14,21 @@
...
 DYN_PEER_PORT=8101
 MEMCACHE_PORT='11211'
+MAX_TOKEN = 4294967295
 
 DEFAULT_DC = 'default_dc'
 
 # gen map of node to random count (3-7) of random tokens (0-MAX_INT)
 token_map = dict()
+token_item = (MAX_TOKEN // (len(sys.argv) -1))
```
timiblossom: Could you please explain a little bit more on this calculation?  I feel like we are missing a for loop here.

balajivenki: Basically here I divide 4294967295 by the number of nodes given for the rack. The script should be run for all nodes in the rack. So if we have 3 racks then I divide 4294967295 / 3 which is 1431655765. And for each of the node while building the conf for it I multiply that with i which is in this case 1, 2, 3. So first node will get 1431655765 and second 2863311530 and third 4294967295. Since we round it I validate that the value cannot be more than 4294967295

scripts/dynomite/generate_yamls.py
```diff
@@ -14,19 +14,21 @@
...
 for i in range(1, len(sys.argv)):
     node = sys.argv[i]
-    token_cnt = 1 #random.randrange(2,7, 1)
     tokens = []
```
timiblossom: Can you please clean out this array?  It was there because we needed it in the original version.  Having it in the new version is unnecessary.

scripts/dynomite/generate_yamls.py
timiblossom: Actually, the way Dynomite works is each rack would own all the keys which have hash values in the range (0, 4294967295).  If you have 3 racks, each would own the same set of data.  A node in a rack might have the same token as another node in another rack.  Hence, in your calculation, you should not build a relationship between the racks by an equidistance or whatsoever.

balajivenki: @timiblossom AFAIK the script should be run per rack and given all the set of nodes in a rack the range(4294967295) is then divided by the number of nodes given. The old script would just print a random number between 0-4294967295. So the script should be used to generate conf per rack.

timiblossom: Please update the description on the top of the file according to what you just mentioned that this script should be run once per rack, and repeat the same step on other racks if needed.  Also, regarding to line 24 on using "tokens = []", using array is not necessary when there is only one element per loop unless I am wrong here.  Then, I think we should be good to commit.  Thanks.

balajivenki: @timiblossom All PR recommendation are done and committed. Please review and it and merge it at your convenience.

PR #351, opened 2016-09-30
Add keepalive to connections

shailesh33:


src/dyn_connection.c
```diff
@@ -92,6 +92,7 @@
  * TODOs: Minh: add explanation for peer-to-peer communication
  */
 
+#define DYN_KEEPALIVE_INTERVAL_S 15 /* seconds */
```
ipapapa: what is the rationale behind 15 seconds?

PR #360, opened 2016-10-13
Fix mset

shailesh33:
o Fix mset command in Dynomite
    o Add test cases for mget and mset

### DIFF
--- a/src/dyn_message.c
+++ b/src/dyn_message.c
@@ -626,7 +626,7 @@ msg_dump(struct msg *msg)
         uint8_t *p, *q;
         long int len;
 
-        p = mbuf->pos;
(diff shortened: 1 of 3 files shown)


src/dyn_message.c
```diff
@@ -626,7 +626,7 @@ msg_dump(struct msg *msg)
         uint8_t *p, *q;
         long int len;
 
-        p = mbuf->start;
+        p = mbuf->pos;
```
ipapapa: Are you sure about that? Seems a major change across all the commands that have been running fine for some time now.

shailesh33:
It's only dumpster my, debugging. Ok I will revert that

PR #373, opened 2016-11-28
Hotfix/v0.5.8 2 local request by pass

shailesh33:
If the message is local, by pass all other logic of key hashing or consistency and just forward the request to local data storage


src/dyn_client.c
```diff
@@ -901,6 +901,14 @@ req_forward(struct context *ctx, struct conn *c_conn, struct msg *msg)
         }
     }
 
+    if (msg->msg_type) {
```
ipapapa: Shouldn't this be `is_local`?

PR #374, opened 2016-11-29
Hotfix/v0.5.8 2 local request by pass

shailesh33:
If the message is local, by pass all other logic of key hashing or consistency and just forward the request to local data storage


ipapapa:
LGTM

PR #375, opened 2016-11-29
Further mset fixes

shailesh33:
o redis post coalesce should take into account error messages for
fragments and act accordingly.
    o While returning error, proper error should be returned to a
fragmented request
    o Some asserts are no longer valid. Removing them
    o msg_send_chain can have a list of requests with no response data when all
the requests are remaining fragments of a already responded request. In
such case it is ok to have nsend = 0
    o In rsp_make_error we were not grabbing the error code of the
current message which could potentially be the first fragment of the
larger request

### DIFF
--- a/src/dyn_client.c
+++ b/src/dyn_client.c
@@ -252,9 +252,6 @@ client_close(struct context *ctx, struct conn *conn)
 static rstatus_t
 client_handle_response(struct conn *conn, msgid_t reqid, struct msg *rsp)
 {
-    ASSERT_LOG(!rsp->peer, "response %lu:%lu has peer set",
-               rsp->id, rsp->parent_id);
-
     // now the handler owns the response.
     ASSERT(conn->type == CONN_CLIENT);
     // Fetch the original request
@@ -700,6 +697,7 @@ remote_req_forward(struct context *ctx, struct conn *c_conn, struct msg *msg,
         // All other cases return a response
         struct msg *rsp = msg_get(c_conn, false, __FUNCTION__);
         msg->done = 1;
+        rsp->type = MSG_RSP_REDIS_ERROR;
         rsp->error = msg->error = 1;
         rsp->err = msg->err = (p_conn ? PEER_HOST_NOT_CONNECTED : PEER_HOST_DOWN);
         rsp->dyn_error = msg->dyn_error = (p_conn ? PEER_HOST_NOT_CONNECTED:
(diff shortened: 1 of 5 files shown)


src/dyn_client.c
```diff
@@ -700,6 +697,7 @@ remote_req_forward(struct context *ctx, struct conn *c_conn, struct msg *msg,
         // All other cases return a response
         struct msg *rsp = msg_get(c_conn, false, __FUNCTION__);
         msg->done = 1;
+        rsp->type = MSG_RSP_REDIS_ERROR;
```
ipapapa: why do we need a type ERROR here if the boolean `rsp->error = 1` is in the next line? Doesn't this sound duplicate code. Can we have an if statement that if it is error to generate an errored response?

src/proto/dyn_redis.c
```diff
@@ -2553,15 +2554,22 @@ redis_pre_coalesce(struct msg *r)
         }
         break;
 
+    case MSG_RSP_REDIS_ERROR:
+        pr->error = r->error;
```
ipapapa: can we have better names in another PR?

src/dyn_client.c
shailesh33: yes I will see if a cleanup is required or possible. I think we still need REDIS_ERROR type since it defines the error. We need a redis error code and a dynomite error code. So not exactly sure what will be the cleanup like. But I know things are a bit confusing here. I will see what can be done next!

src/dyn_dnode_client.c
```diff
@@ -528,7 +528,6 @@ dnode_rsp_send_done(struct context *ctx, struct conn *conn, struct msg *rsp)
 
     ASSERT(!rsp->request && req->request);
     ASSERT(req->selected_rsp == rsp);
-    ASSERT(req->done && !req->swallow);
```
shailesh33: because the request may or may not be marked for swallow. And this function is called after the response is sent out, so obviously it is done.

src/proto/dyn_redis.c
shailesh33: yes

src/dyn_client.c
ipapapa: IMHO, we probably need a REDIS_ERROR but this does not need to be carried in the `rsp->type` but rather just be used when we respond with an error. Anyway, obviously the code might be confusing to clean it up so no worries just good to have.

PR #382, opened 2016-12-01
More renames and correct checks for peer field

shailesh33:
* rename to req and rsp whereever applicable
* req always uses selected_rsp
* rsp always uses peer

### DIFF
--- a/src/dyn_core.c
+++ b/src/dyn_core.c
@@ -387,20 +387,20 @@ static void
 core_timeout(struct context *ctx)
 {
 	for (;;) {
-		struct msg *msg;
+		struct msg *req;
 		struct conn *conn;
 		msec_t now, then;
 
-		msg = msg_tmo_min();
-		if (msg == NULL) {
+		req = msg_tmo_min();
+		if (req == NULL) {
 			ctx->timeout = ctx->max_timeout;
 			return;
 		}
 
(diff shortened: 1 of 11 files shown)


src/dyn_core.c
```diff
@@ -387,20 +387,20 @@ static void
 core_timeout(struct context *ctx)
 {
 	for (;;) {
-		struct msg *msg;
+		struct msg *req;
```
ipapapa: Does this mean that whenever we use `req` it will imply `msg` even outside of this function? or would `req_msg` be clearer?

shailesh33: req always means msg as far as I have seen and understood the code.

src/dyn_core.c
```diff
@@ -419,10 +419,10 @@ core_timeout(struct context *ctx)
 			return;
 		}
 
-        log_warn("req %"PRIu64" on %s %d timedout, timeout was %d", msg->id,
-                 conn_get_type_string(conn), conn->sd, msg->tmo_rbe.timeout);
+        log_warn("req %"PRIu64" on %s %d timedout, timeout was %d", req->id,
```
shailesh33: the entire file has tabs in the beginning. Please ignore this one, I will make a seperate checkin to fix the tabs

PR #383, opened 2016-12-03
Remove the concept of distribution

shailesh33:
* We use only one type of distribution thats provided by the dynomite
  manager to keep things simple. Cleanup code related to that
* Reduce some warnings

### DIFF
--- a/src/dyn_vnode.h
+++ b/src/dyn_vnode.h
@@ -1,0 +1,6 @@
+#pragma once
+#include <dyn_types.h>
+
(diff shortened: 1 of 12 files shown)


src/dyn_vnode.h
```diff
@@ -0,0 +1,6 @@
+#pragma once
```
ipapapa: the name vnode is a little confusing as we are not really having vnodes.

shailesh33: that will be a seperate PR

PR #385, opened 2016-12-08
enable_gossip property in the .yml

ipapapa:
Simplifying configuration commands by moving gossip enable/disable to YAML.


ipapapa:
Replaced by #386

PR #388, opened 2016-12-13
Implement generic task scheduler

shailesh33:
* Implement a task scheduler that takes tasks with 1 argument for now
* Create a reconnect task everytime a connection to peer is broken
* This reconnect task gets executed in the main loop
* Individual peers will have their own backoff mechanism to be followed
* Convert the connection retry timeouts to msec


shailesh33:
@ipapapa note that I tested the travis in a loop and not once I received a "Peer node not connected" error. 
However in a cluster (dyno_sh_dev) when I arbitrarily killed a dynomite process, I was expecting all other nodes to keep retrying connection to the dead node every 10 sec (which I think might be a logging overkill, and propose to increase it to 30 secs with exponential backoff) but that didnt happen. The problem is we do not handle recv() returning 0 correctly. I will rectify that in a next PR since I do not want to make this PR too huge and complicated.

src/dyn_core.c
```diff
@@ -610,12 +606,14 @@ core_loop(struct context *ctx)
 
 	core_process_messages();
 
-	nsd = event_wait(ctx->evb, ctx->timeout);
-	if (nsd < 0) {
-		return nsd;
-	}
+    core_timeout(ctx);
+    execute_expired_tasks(0);
+    ctx->timeout = MIN(ctx->timeout, time_to_next_task());
```
ipapapa: what does `MIN` do?

src/dyn_dnode_peer.c
```diff
@@ -1009,11 +1017,10 @@ dnode_peer_connected(struct context *ctx, struct conn *conn)
 
     conn->connecting = 0;
     conn->connected = 1;
+    peer->state = NORMAL;
 
-        if (log_loggable(LOG_INFO)) {
-       log_debug(LOG_INFO, "dyn: peer connected on sd %d to server '%.*s'", conn->sd,
-              server->endpoint.pname.len, server->endpoint.pname.data);
-        }
+    log_notice("connected to %s '%.*s' from sd %u", conn_get_type_string(conn),
```
ipapapa: What did we change `log_debug` to `log_notice`? First time I see the use of `log_notice`.

src/dyn_dnode_peer.c
```diff
@@ -1009,11 +1017,10 @@ dnode_peer_connected(struct context *ctx, struct conn *conn)
 
     conn->connecting = 0;
     conn->connected = 1;
+    peer->state = NORMAL;
```
ipapapa: why are we changing the state here? Is `peer->state` used for gossip?

src/dyn_dnode_peer.c
```diff
@@ -59,6 +61,12 @@ dnode_peer_unref(struct conn *conn)
     peer = conn->owner;
     conn->owner = NULL;
 
+    /* FIXME: This is a wrong place to schedule a reconnection since this unref has no
```
ipapapa: Is there going to be another PR that will fix this?

src/dyn_core.c
shailesh33: gets the minimum of the two args

src/dyn_dnode_peer.c
shailesh33: there was a log for "connecting to ...." but no log for "connected to ...." so I added that.
log_notice is level 4 I think which is the default level.

src/dyn_dnode_peer.c
shailesh33: havent planned but there are other low hanging fruits to cut and eat.

src/dyn_dnode_peer.c
shailesh33: we marked the peer as down in dnode_peer_failure, but when we are connected we didnt mark it as NORMAL, so although the connection was established, we were not passing down the data because we had marked the node as DOWN. So making it NORMAL as soon as it is connected

PR #389, opened 2016-12-13
Migrate the wiki to gh-pages and add mkdocs

diegopacheco:
@ipapapa

This is the very first version using mkdocs(http://www.mkdocs.org/)

You can access my fork on: http://diegopacheco.github.io/dynomite/

Once you accept the PR: DM wiki will be here: http://netflix.github.io/dynomite/

On the repo we have the site which was generated by mkdocs via $ mkdocs build. The source code in markdown is on the folder: mkdocs-site.

Cheers,
Diego Pacheco


ipapapa:
LGTM, it is just porting the wiki to a website. We probably need to follow up updating our README and related links to point to the website and then delete all content from wiki. @shailesh33 thoughts?

ipapapa:
@diegopacheco if we want to update the website, how easy would that be? Moreover, I am looking into the structure and it is not the same as the wiki which makes it hard to parse it. Can we follow the wiki structure?

diegopacheco:
@ipapapa 

Sure. I will try to submit something before Xmas.

Cheers,
Diego Pacheco

diegopacheco:
@ipapapa 

I made the change you requested. I used prefixed numbers so I can grantee the same order as the wiki.

ipapapa:
Thanks!

PR #391, opened 2016-12-14
Fix autoeject for datastore

shailesh33:
* when a datastore is autoejected, we never got it back. Fix it
* Add new dyn_error_code to indicate a dynomite error to the higher layer

### DIFF
--- a/src/dyn_client.c
+++ b/src/dyn_client.c
@@ -413,12 +413,12 @@ send_rsp_integer(struct context *ctx, struct conn *c_conn, struct msg *req)
 
 static void
 req_forward_error(struct context *ctx, struct conn *conn, struct msg *req,
-                  err_t err)
+                  err_t error_code, err_t dyn_error_code)
 {
     if (log_loggable(LOG_INFO)) {
        log_debug(LOG_INFO, "forward req %"PRIu64" len %"PRIu32" type %d from "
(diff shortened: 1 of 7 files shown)


src/dyn_client.c
```diff
@@ -979,6 +980,9 @@ msg_local_one_rsp_handler(struct msg *req, struct msg *rsp)
                rsp->parent_id);
     req->awaiting_rsps = 0;
     rsp->peer = req;
+    req->is_error = rsp->is_error;
+    req->error_code = rsp->error_code;
```
ipapapa: two error codes???

shailesh33: One is errno, system error code and other is dynomite error code. Historical reasons

src/dyn_message.h
```diff
@@ -210,7 +210,8 @@ typedef enum msg_type {
 
 
 typedef enum dyn_error {
-    UNKNOWN_ERROR,
+    DYNOMITE_UNKNOWN_ERROR,
```
shailesh33: Good catch. I will make it Dynomite_no_quorum_achieved

src/dyn_client.c
ipapapa: maybe rename `error_code` to something more representative in the future.

PR #393, opened 2016-12-19
Single connection to datastore and peer

shailesh33:
* Having multiple connections to server and spraying messages will corrupt data, unless there is a good mapping of client connections to datastore connections etc
* Get ride of server_connections option in YAML (deprecated)
* rename dyn_state to dyn_parse_state where appropriate

### DIFF
--- a/src/dyn_core.h
+++ b/src/dyn_core.h
@@ -132,12 +132,12 @@ typedef enum dyn_state {
 	WRITES_ONLY = 2,
 	RESUMING    = 3,
 	NORMAL      = 4,
-	SUSPENDING  = 5,
-	LEAVING     = 6,
+	//SUSPENDING  = 5,
+	//LEAVING     = 6,
 	JOINING     = 7,
 	DOWN        = 8,
-	REMOVED     = 9,
-	EXITING     = 10,
+	//REMOVED     = 9,
+	//EXITING     = 10,
 	RESET       = 11,
 	UNKNOWN     = 12
 } dyn_state_t;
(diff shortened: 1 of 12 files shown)


src/dyn_core.h
```diff
@@ -132,12 +132,12 @@ typedef enum dyn_state {
 	WRITES_ONLY = 2,
 	RESUMING    = 3,
 	NORMAL      = 4,
-	SUSPENDING  = 5,
-	LEAVING     = 6,
+	//SUSPENDING  = 5,
```
ipapapa: Not using these states for gossip in the future?

src/dyn_dnode_peer.c
```diff
@@ -917,8 +888,8 @@ dnode_peer_replace(void *rmsg)
 
 
     if (s != NULL) {
-        log_debug(LOG_INFO, "Found an old node to replace '%.*s'", s->name.len, s->name.data);
-        log_debug(LOG_INFO, "Replace with address '%.*s'", node->name.len, node->name.data);
+        log_debug(LOG_NOTICE, "Found an old node to replace '%.*s'", s->name.len, s->name.data);
```
ipapapa: I never understood why we need `loga(LOG_INFO,...)` and `log_debug(LOG_NOTICE,...`. It would have been better to have a wrapper function that makes it consistent.

src/dyn_server.c
```diff
@@ -37,14 +37,13 @@ server_ref(struct conn *conn, void *owner)
 
     ASSERT(conn->type == CONN_SERVER);
 	ASSERT(conn->owner == NULL);
+	ASSERT(server->conn == NULL);
```
ipapapa: The indentation here is a little off.

src/dyn_core.h
shailesh33: We are currently not using them, that is why I just commented to reduce confusion.

src/dyn_dnode_peer.c
```diff
@@ -38,8 +38,8 @@ dnode_peer_ref(struct conn *conn, void *owner)
     conn->addr = peer->endpoint.addr;
     string_duplicate(&conn->pname, &peer->endpoint.pname);
 
-    peer->ns_conn_q++;
-    TAILQ_INSERT_TAIL(&peer->s_conn_q, conn, conn_tqe);
+    ASSERT(peer->conn == NULL);
```
shailesh33: we use this in travis:

    CFLAGS="-ggdb3 -O0" autoreconf -fvi && ./configure --enable-debug=log && make

Is that what you mean by debug? or --enable-debug=full?

src/dyn_dnode_peer.c
shailesh33: Honestly, I do not know why we need so many (12) levels of logging. I created these wrapper functions for different levels as and how needed. The other log_debug function is inherited.

src/dyn_dnode_peer.c
shailesh33: I just ran the test. THere was a issue in the stats code which I fixed.

PR #397, opened 2016-12-21
Start peer with state DOWN initially

shailesh33:
* Start with peer as DOWN initially and then make them NORMAL after connection is established.
* A NORMAL state means a node is up and connected.
* Self node is NORMAL by default.
* Peer node becomes normal after a connection is successful.
* Also make sure the peer is marked DOWN as soon as the connection breaks so that the higher layers can react accordingly.
* This will help in rerouting traffic to a different rack as soon as the node in the remote region goes down

### DIFF
--- a/src/dyn_conf.c
+++ b/src/dyn_conf.c
@@ -213,7 +213,7 @@ conf_seed_each_transform(void *elem, void *data)
     s->owner = NULL;
     s->endpoint.pname = cseed->pname;
 
-    s->state = NORMAL;//assume peers are normal initially
+    s->state = DOWN;//assume peers are down initially
 
     uint8_t *p = cseed->name.data + cseed->name.len - 1;
     uint8_t *start = cseed->name.data;
--- a/src/dyn_core.c
+++ b/src/dyn_core.c
@@ -33,11 +33,54 @@
 #include "event/dyn_event.h"
 #include "dyn_task.h"
 
+static void
+core_print_peer_status(void *arg1)
+{
+    struct context *ctx = arg1;
+    struct server_pool *sp = &ctx->pool;
+    // iterate over all peers
+    uint32_t dc_cnt = array_n(&sp->datacenters);
+    uint32_t dc_index;
+    for(dc_index = 0; dc_index < dc_cnt; dc_index++) {
+        struct datacenter *dc = array_get(&sp->datacenters, dc_index);
+        if (!dc)
+            log_panic("DC is null. Topology not inited proerly");
+        uint8_t rack_cnt = (uint8_t)array_n(&dc->racks);
+        uint8_t rack_index;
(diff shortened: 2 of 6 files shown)


src/dyn_core.c
```diff
@@ -33,11 +33,54 @@
...
+        uint8_t rack_index;
+        for(rack_index = 0; rack_index < rack_cnt; rack_index++) {
+            struct rack *rack = array_get(&dc->racks, rack_index);
+            uint8_t i = 0;
+            for (i = 0; i< rack->ncontinuum; i++) {
+                struct continuum *c = &rack->continuum[i];
+                uint32_t peer_index = c->index;
+                struct node *peer = array_get(&sp->peers, peer_index);
+                if (!peer)
+                    log_panic("peer is null. Topology not inited proerly");
```
ipapapa: so we are doing `log_panic` and then subsequently `log_notice`? Shouldn't both be the same?

shailesh33:
Log panic will exit the process. So it's more like.assert. But always on.

src/dyn_core.c
shailesh33: Log panic will exit the process. So it's more like.assert. But always on.

shailesh33: log_panic is if !peer

src/dyn_conf.c
```diff
@@ -213,7 +213,7 @@ conf_seed_each_transform(void *elem, void *data)
     s->owner = NULL;
     s->endpoint.pname = cseed->pname;
 
-    s->state = NORMAL;//assume peers are normal initially
+    s->state = DOWN;//assume peers are down initially
```
ipapapa: so we are changing the status to DOWN here from NORMAL. But where are we changing back to NORMAL?

shailesh33: Initially its DOWN. WHen the connection is established, the state is changed to NORMAL, 
look for dnode_peer_connected

src/dyn_core.c
```diff
@@ -33,11 +33,54 @@
...
+    peer->state = state;
+}
+
 static rstatus_t
 core_init_last(struct context *ctx)
 {
     core_debug(ctx);
     preselect_remote_rack_for_replication(ctx);
+    // Print the network health once after 30 secs
+    schedule_task_1(core_print_peer_status, ctx, 30000);
```
ipapapa: The more tasks we add the more threads we will create. Obviously, these are lightweight, but we do perform core_affinity and allocate two cores to Dynomite. Will this cause any issue if we create many scheduled tasks?

shailesh33: This is not a "task" as you know in java. Its just a struct with 1 argument and 1 function pointer. The main thread will execute each of them. No Seperate thread

Issue #96, opened 2015-03-15
default config fails to start when pem file not found


miketheman:
Introduced in #57.

The [default config](https://github.com/Netflix/dynomite/blob/438b6dbbb49fbca64897d46badcc7dc1562da7b7/conf/dynomite.yml) has no configuration set for crypto, yet it attempts to load it by default, and fails when not found.

```
root@4622f90515db:~# pwd
/root
root@4622f90515db:~# which dynomite
/usr/local/sbin/dynomite
root@4622f90515db:~# dynomite -c dynomite/conf/dynomite.yml
[Sun Mar 15 16:17:35 2015] dynomite.c:194 dynomite-0.1.19 built for Linux 3.18.5-tinycore64 x86_64 started on pid 8668
[Sun Mar 15 16:17:35 2015] dynomite.c:199 run, rabbit run / dig that hole, forget the sun / and when at last the work is done / don't sit down / it's time to dig another one
[Sun Mar 15 16:17:35 2015] dyn_crypto.c:45 Error: file conf/dynomite.pem not exists
[Sun Mar 15 16:17:35 2015] dyn_core.c:87 Failed to initialize crypto!!!
[Sun Mar 15 16:17:35 2015] dynomite.c:205 done, rabbit done
root@4622f90515db:~#
```

The error message comes from https://github.com/Netflix/dynomite/blob/438b6dbbb49fbca64897d46badcc7dc1562da7b7/src/dyn_crypto.c#L44-L47 - which makes sense, as the default-set path of `conf/dynomite.pem` is not reachable from the current path, however this should not be loaded until a `pem_key_file` setting is added in the config.

This is likely due to the definition here: https://github.com/Netflix/dynomite/blob/438b6dbbb49fbca64897d46badcc7dc1562da7b7/src/dyn_conf.h#L86 which is inherited by all configurations, even when unspecified.

I _think_ this shouldn't be declared as a constant, rather some empty value that does not attempt to initialize crypto on startup when the parameter is not set.

timiblossom:
This makes sense.  Will fix this. Thanks.

nimeops:
Hi.
Seems to be always broken??

nimeops:
hi,
something i don't understand.

On several servers, no problem if there is no crypto (no pem file), and others, we've got following error:
 
load_private_rsa_key_by_file:44 Error: file conf/dynomite.pem does not exist
[...]

PR #398, opened 2016-12-23
making loading PEM file optional if not provided

ipapapa:
Addresses: https://github.com/Netflix/dynomite/issues/96

### DIFF
--- a/src/dyn_conf.c
+++ b/src/dyn_conf.c
@@ -2286,19 +2286,24 @@ conf_validate_pool(struct conf *cf, struct conf_pool *cp)
         log_debug(LOG_INFO, "setting env to default value:%s", CONF_DEFAULT_ENV);
     }
 
-    if (string_empty(&cp->pem_key_file)) {
-        string_copy_c(&cp->pem_key_file, (const uint8_t *)PEM_KEY_FILE);
-        log_debug(LOG_INFO, "setting pem key file to default value:%s", PEM_KEY_FILE);
-    }
+    /**
+     * if the secure server options is not set, then no need to load KEY files
+     */
+    if (string_compare(&cp->secure_server_option, (const uint8_t *)CONF_STR_NONE) == 0) {
+      if (string_empty(&cp->pem_key_file)) {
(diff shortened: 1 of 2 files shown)


src/dyn_conf.c
```diff
@@ -2286,19 +2286,24 @@ conf_validate_pool(struct conf *cf, struct conf_pool *cp)
...
-        log_debug(LOG_INFO, "setting reconciliation key file to default value:%s", RECON_KEY_FILE);
-    }
-
-    if (string_empty(&cp->recon_iv_file)) {
-        string_copy_c(&cp->recon_iv_file, (const uint8_t *)RECON_IV_FILE);
-        log_debug(LOG_INFO, "setting reconciliation IV file to default value:%s", RECON_IV_FILE);
+    /**
+     * if the secure server options is not set, then no need to load KEY files
+     */
+    if (string_compare(&cp->secure_server_option, (const uint8_t *)CONF_STR_NONE) == 0) {
```
shailesh33: this secure server option different from the reconciliation option right?

ipapapa: Yeah, I thought about that as well but by not setting it at all it will also bypass the issue of recon PEM files not existing. I open to other approaches...

shailesh33: 1) disable starting of recon code unless we are prod ready? 
2) Have a choice to start it or not based on a yaml config just like secure_server_option?

ipapapa: I can do the (2) but what if somebody uses `secure_server_option = datacenter` why would he want a different security setting for the anti-entropy? Effectively the same data are replicated. For example, a common mistake would be `datacenter` in secure_server_option and not use the same in anti-entropy.

For (1) the recon code even if it returns `DN_ERROR` Dynomite prints out a debug message with the failure, and does not start so that is not affected, to that extent it is disabled. I can add a temporal property in YAML that says entropy and it will be used only when somebody wants to use anti-entropy. Thoughts?

shailesh33: we have seperate PEM files for anti entropy right?

ipapapa: yes

PR #401, opened 2016-12-28
Randomize and extend aes test

shailesh33:
Run AES test on 10M buffers generated randomly with random length at
runtime and make sure encrypt decrypt functions are fine.
Note: This does not test the message encryption decryption yet


shailesh33:
I didnt realize that this is a internal function that our message encryption calls. The aes_msg_test is what needs to be extended really. But I was already there so just finished it nevertheless

PR #402, opened 2016-12-31
Better Error handling during request forwarding

shailesh33:
This is an important patch that sort of redesigns some of the APIs used by the coordinator.
The general idea is to propagate errors back so that the higher layers can act accordingly.
The errors are of two types: rstatus which talks about overall operation failure and dyn_error_code which is a more precise error code. Eventually these should be combined.
The Dnode client coordinator (the coordinator) which receives request from peer nodes reuses the client co-ordinator functionality.

### DIFF
--- a/src/dyn_client.c
+++ b/src/dyn_client.c
@@ -87,7 +87,8 @@ client_unref_internal_try_put(struct conn *conn)
     ASSERT(conn->waiting_to_unref);
     unsigned long msgs = dictSize(conn->outstanding_msgs_dict);
     if (msgs != 0) {
-        log_warn("conn %p Waiting for %lu outstanding messages", conn, msgs);
+        log_warn("conn %s %d Waiting for %lu outstanding messages",
+                 conn_get_type_string(conn), conn->sd, msgs);
         return;
     }
     struct server_pool *pool;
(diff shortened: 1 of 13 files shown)


shailesh33:
@ipapapa I purposely didnt combine the commits. Look at individual commits and let me know.

Better approach would be to actually read the code starting from dyn_client.c:req_forward()

shailesh33:
I also tried arbitrarily killing nodes in local region and remote region.

ipapapa:
When you say eventually the error messages should be combined, is this going to be done in a follow-up PR?

src/dyn_client.c
```diff
@@ -651,9 +651,10 @@ admin_local_req_forward(struct context *ctx, struct conn *c_conn, struct msg *re
     local_req_forward(ctx, c_conn, req, key, keylen);
 }
 
-void
+rstatus_t
```
ipapapa: if the higher level function is a `void` what would be the reason to return `rstatus_t` in `remote_req_forward`?

src/dyn_client.c
```diff
@@ -782,8 +782,10 @@ req_forward_all_local_racks(struct context *ctx, struct conn *c_conn,
                      rack_msg->parent_id, rack->name->len, rack->name->data);
 
             s = remote_req_forward(ctx, c_conn, rack_msg, rack, key, keylen, &dyn_error_code);
-            if ((s != DN_OK) && (req->consistency != DC_ONE)) {
-                req_forward_error(ctx, c_conn, req, s, dyn_error_code);
+            if (s != DN_OK) {
```
ipapapa: Have we seen this message leak in production?

shailesh33:
I saw it during my testing.

shailesh33:
Eventually means in one of the follow up PRs but not immediate.

shailesh33:
We use the return code to determine if we want to free the messages or not.

PR #403, opened 2017-01-03
Remote DC replication Failover

shailesh33:
This is the final patch in the series to enable remote DC replication
failover. If a node figures a remote node is marked as DOWN, it tries to
failover to another node. This does not failover those requests which
were already queued on the connection and could not be forwarded due to
network issues, but only those that arrived after the remote peer was
marked as DOWN.

### DIFF
--- a/src/dyn_client.c
+++ b/src/dyn_client.c
@@ -782,7 +782,7 @@ req_forward_remote_dc(struct context *ctx, struct conn *c_conn, struct msg *req,
                       struct mbuf *orig_mbuf, uint8_t *key, uint32_t keylen,
                       struct datacenter *dc)
 {
-    uint32_t rack_cnt = array_n(&dc->racks);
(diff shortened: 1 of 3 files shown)


src/dyn_client.c
```diff
@@ -801,15 +801,47 @@ req_forward_remote_dc(struct context *ctx, struct conn *c_conn, struct msg *req,
              req->id, req->parent_id, rack_msg->id, rack_msg->parent_id);
     rack_msg->swallow = true;
 
-    log_debug(LOG_DEBUG, "forwarding request to conn '%s' on rack '%.*s'",
+    log_debug(LOG_DEBUG, "forwarding request from conn '%s' on rack '%.*s'",
```
ipapapa: Do we want to have `from` and `to` if we have the outgoing information?

ipapapa:
Overal LGTM

PR #407, opened 2017-01-05
Custom printf specifier

shailesh33:
Implemented a printf handler of M to print out objects.
Created a new object type enum for each type of object. Right now it is
only for request, response and connection.
Printing information about this objects now is very simple is
customizable at one place.

This, I believe will ease printing information in the logs and hence help debugging issues


shailesh33:
A code like this:
log_debug(LOG_DEBUG, "%M received %M key '%.*s' adding to dict", c_conn,
              req, keylen, key);

will print:

[2017-01-05 13:44:12.122] req_forward:888 <conn 0x21da770 CLIENT 13> received <req 0x21dc4c0 1:0 REQ_REDIS_GET> key 'test' adding to dict

PR #409, opened 2017-01-18
Docker

v1r3n:
* Separate startup script to start redis and dynomite

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -36,12 +36,13 @@ RUN apt-get update && apt-get install -y \
 	redis-server \
 	tcl8.5
 
-# Get Redis Running
-RUN service redis-server start
-
 # Clone the Dynomite Git
 RUN git clone https://github.com/Netflix/dynomite.git
 RUN echo 'Git repo has been cloned in your Docker VM'
+
+RUN echo 'copy file'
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -71,16 +72,9 @@ EXPOSE 22222
 RUN echo 'Exposing client port for Dynomite 8102'
 EXPOSE 8102
```
ipapapa: We don't really need to `EXPOSE 22122` but shall we add it for debugging purposes?

docker/scripts/startup.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+#Start redis server on 22122
+redis-server --port 22122 &
+
+src/dynomite --conf-file=conf/redis_single.yml -v11
```
ipapapa: any reason why just don't let it start with default YAML. Any specific reason on selecting `redis_single.yml`?

v1r3n: I used the same command from the existing one.  Let me try with the defaults.

docker/Dockerfile
v1r3n: 22122 is already added.

PR #410, opened 2017-01-19
listen on all the available interfaces (not just on 127.0.0.1)

v1r3n:
Updated the listen address to use all the available interfaces.

### DIFF
--- a/conf/redis_single.yml
+++ b/conf/redis_single.yml
@@ -1,8 +1,8 @@
 dyn_o_mite:
-  dyn_listen: 127.0.0.1:8101
+  dyn_listen: 0.0.0.0:8101
   data_store: 0
-  listen: 127.0.0.1:8102
-  dyn_seed_provider: simple_provider 
+  listen: 0.0.0.0:8102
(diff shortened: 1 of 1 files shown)


conf/redis_single.yml
```diff
@@ -1,8 +1,8 @@
 dyn_o_mite:
-  dyn_listen: 127.0.0.1:8101
+  dyn_listen: 0.0.0.0:8101
   data_store: 0
-  listen: 127.0.0.1:8102
-  dyn_seed_provider: simple_provider 
+  listen: 0.0.0.0:8102
+  dyn_seed_provider: simple_provider
   servers:
   - 127.0.0.1:22122:1
```
shailesh33: why this is not 0.0.0.0?

v1r3n: Only listen needed to be changed to bind the the listening socket. 
servers part, I think it used to by dynomite to connect to local redis instance and 127.0.0.1 works just fine.

shailesh33: oh yes!! my bad

PR #420, opened 2017-02-13
stats: add nesting around servers for easier parsing

jbdalido:
Adding a simple nesting with key `servers` for easier parsing of the json delivered by `/info` on the api.


shailesh33:
I am not exactly sure if this affects the dynomite-manager to collect stats. I will have to verify that.

jbdalido:
For now the `/info` is mostly unusable for clean parsing, we are building an external prometheus exporter for a production use-case and it could be great to offer a cleaner output. Thanks a lot

PR #440, opened 2017-04-28
Connection cleanup

shailesh33:
* Move static internal functions to dyn_connection_internal.c
* Make object printing more generic
* reduce 4 conn_get_* apis into one conn_get
* Move cleanup_charptr to a .c and reduce warnings by 50

### DIFF
--- a/src/dyn_connection_internal.c
+++ b/src/dyn_connection_internal.c
@@ -4,9 +4,6 @@
  */ 
 
 /*
- * twemproxy - A fast and lightweight proxy for memcached protocol.
- * Copyright (C) 2011 Twitter, Inc.
- *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
--- a/src/dyn_connection_internal.c
+++ b/src/dyn_connection_internal.c
@@ -1,0 +1,223 @@
+/*
+ * Dynomite - A thin, distributed replication layer for multi non-distributed storages.
+ * Copyright (C) 2014 Netflix, Inc.
+ */ 
+
+/*
+ * twemproxy - A fast and lightweight proxy for memcached protocol.
+ * Copyright (C) 2011 Twitter, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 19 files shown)


src/dyn_connection_internal.c
```diff
@@ -0,0 +1,256 @@
+/*
+ * Dynomite - A thin, distributed replication layer for multi non-distributed storages.
+ * Copyright (C) 2014 Netflix, Inc.
+ */ 
+
+/*
+ * twemproxy - A fast and lightweight proxy for memcached protocol.
```
ipapapa: This is a new function that does not exist in Twemproxy. We do not need this license anymore.

src/dyn_types.c
```diff
@@ -0,0 +1,7 @@
+#include "dyn_types.h"
```
ipapapa: Are you planning to expand `dyn_types.c`?

src/dyn_message.c
```diff
@@ -162,17 +162,21 @@ static struct string msg_type_strings[] = {
...
     return fprintf(stream, "<REQ %p %lu:%lu %.*s>", req, req->id, req->parent_id,
                    req_type->len, req_type->data);
 }
 
-int
-print_rsp(FILE *stream, struct msg *rsp)
+static int
+print_rsp(FILE *stream, const struct object *obj)
 {
+    ASSERT(obj->type == OBJ_REQ);
```
ipapapa: It is nice to add the `ASSERT` but are they actually working? In the past debug had some issues.

shailesh33: asserts are working with the new build process that we did in our jenkins job. but I will check

src/dyn_types.c
shailesh33: Depends. not sure.

src/dyn_connection_internal.c
shailesh33: ok

src/dyn_types.c
ipapapa: then why did you create a new file just for migrating a small function?

shailesh33: yes for now. I didnt want to put in dyn_core but I will revisit it.

PR #451, opened 2017-05-29
Update dynomite.c

asgs:
Converting infinite for loop to a deterministic while loop, while parsing command line arguments.


shailesh33:
Thanks.

PR #453, opened 2017-06-15
Connection pool

shailesh33:
This PR adds support for multiple connections using a connection pool.
There are 3 new entries to the conf: datastore_connections, local_peer_connections, remote_peer_connectons. The default is 1. The datastore and a node object now creates a connection pool instead of connection. The connection pool internally creates the required number of connections. When the higher layer has to send a message it requests a connection from the connection from the connection pool and uses it to forward the message.

There is a non-trivial interactions between the higher layer and the connection pool because of the way event loop works. The main thread is the one that runs the event loop so it gets the notification of when a connection gets disconnected etc. Hence the main thread has to notify the connection pool that the connection is errorred or closed. Hence the connection pool has  _notify_* apis. This is a little counter intuitive but the next logical step for connection pooling.

The conn_pool_get is the mapping function which maps a tag (higher level socket id for now) with the underlying connection in the pool. The connection pool should guarantee the same underlying connection for a given tag, that way the ordering of messages is maintained. This mapping function can be changed to try various techniques like having dedicated connection to datastore from remote peer, local peer and local clients etc.


ipapapa:
There are a few conflicts that need to be resolved manually.

src/dyn_conf.h
```diff
@@ -61,16 +61,14 @@ struct conf_listen {
  * Server configuration.
  */
 struct conf_server {
-    struct string   pname;       /* server: as "name:port:weight" */
-    struct string   name;        /* name */
+    struct string   pname;       /* server: as "name:port:weight" or "hostname:port:rack:dc:tokens" */
```
ipapapa: Do we support `hostname` or just IP addresses?

conf/a_dc1.yml
```diff
@@ -15,3 +15,6 @@ dyn_o_mite:
   pem_key_file: conf/dynomite.pem
   data_store: 0
   stats_listen: 0.0.0.0:22221
+  datastore_connections: 3
+  local_peer_connections: 3
```
shailesh33: Havent seen improvements with redis obviously. Didnt try on ARDB yet.
I will make the changes for YAML

src/dyn_gossip.c
```diff
@@ -954,6 +954,14 @@ gossip_pool_init(struct context *ctx)
             string_deinit(&gnode->name);
             string_copy(&gnode->name, b_address, dn_strlen(b_address));
         } else {
+
+            char *local_ip4 = hostname_to_private_ip4( (char *) gnode->name.data);
```
shailesh33: Initially, here, we were using the hostnames, and later from florida we were taking private ip. So what would happen is, we start and create a map of hostname -> node. But the first pull from Florida of get seeds we would convert hostname to ips and try to find the IP in the map. Obviously we didnt find and we complained that the node is replaced and we would disconnect and reconnect. To avoid this disconnect and reconnect, we are keeping the private IP in the map right from the first step. THis avoids the reconnection logic to be kicked in unnecessarily. 
Now why private ip and not public IP, I do not know exactly. But I did this to keep it in tune with the current code

src/dyn_conf.h
shailesh33: Not exactly sure of it.

PR #454, opened 2017-06-15
Reduce malloc() and free() call when call dnode_peer_for_key_on_rack()

lampmanyao:


src/dyn_token.h
```diff
@@ -12,7 +12,7 @@
 
 struct dyn_token {
     uint32_t signum;
-    uint32_t *mag;
+    uint32_t mag[4];
```
shailesh33: I have not looked at this code for a while. What do we store in mag? is 4 enough?

src/dyn_token.c
```diff
@@ -35,16 +35,12 @@ void
 init_dyn_token(struct dyn_token *token)
 {
 	token->signum = 0;
-	token->mag = NULL;
```
shailesh33: Initialize mag here?

shailesh33:
Overall it looks good. thanks a lot.

src/dyn_token.h
lampmanyao: We store dnode's token or hash value of key in mag.
All of hash functions use 4 bytes, except hash_murmur3(), which uses 128-bit storage.
I think 4 is enough.

shailesh33: Appreciate. Since you are at it, can you ASSERT when it goes beyond 4 bytes?

Issue #220, opened 2015-12-22
Memcached requests are failing


proger:
When testing with a [modified version](https://gist.github.com/proger/aeb0f665fe9d092c7256) of a docker image:

version info:

```
# docker run -it --rm -P dyno
[2015-12-22 15:45:39.902] dn_print_run:204 dynomite-0.5.6 built for Linux 3.19.7 x86_64 started on pid 12
```

tested with (similar results for get/put):

```
# (printf "quit\r\n"; sleep 3) | nc localhost 8102
```

log snippet:

```
[2015-12-22 15:47:38.732] event_wait:263 epoll 0001 triggered on conn 0x13bbcd0
[2015-12-22 15:47:38.733] core_core:397 event 00FF on PROXY 7
[2015-12-22 15:47:38.734] _dn_alloc:253 malloc(88) at 0x13c4310 @ dyn_dict.c:176
[2015-12-22 15:47:38.734] client_ref:100 ref conn 0x13c4140 owner 0x13b22a0 into pool 'dyn_o_mite'
[2015-12-22 15:47:38.735] conn_get:380 get conn 0x13c4140 CLIENT
[2015-12-22 15:47:38.735] stats_pool_to_metric:1519 metric 'client_connections' in pool 0
[2015-12-22 15:47:38.735] _stats_pool_incr:1538 incr field 'client_connections' to 0
[2015-12-22 15:47:38.736] event_add_conn:206 adding conn 0x13c4140(CLIENT) to active
[2015-12-22 15:47:38.736] proxy_accept:355 accepted CLIENT 10 on PROXY 7 from '127.0.0.1:37055'
[2015-12-22 15:47:38.736] proxy_accept:297 accept on PROXY 7 not ready - eagain
[2015-12-22 15:47:38.737] stats_swap:1436 skip swap of current 0x13b2130 shadow 0x13b41a0 as aggregator is busy
[2015-12-22 15:47:38.737] core_process_messages:504 length of C2G_OutQ : 0
[2015-12-22 15:47:38.737] event_wait:263 epoll 0004 triggered on conn 0x13c4140
[2015-12-22 15:47:38.737] core_core:397 event FF00 on CLIENT 10
[2015-12-22 15:47:38.738] event_del_out:180 removing conn 0x13c4140(CLIENT) from active
[2015-12-22 15:47:38.738] stats_swap:1436 skip swap of current 0x13b2130 shadow 0x13b41a0 as aggregator is busy
[...]

PR #474, opened 2017-09-12
Multikey operation optimization

shailesh33:
This is a major change in the multi key operations like `MGET`, `MSET` and `DEL`.
It significantly reduces the latencies and message overhead by smartly splitting the multiple key requests across nodes. 
Basis:
For example in a rack with 3 nodes, and a request like `mget k1 k2 k3 k4 k5 k6 k7 k8 k9`, today, Dynomite would split the requests into 9 requests of `mget k1, mget k2`.... etc, one for each key. 
With the new changes, the request will be split at the max into as many requests as the number of nodes in the rack like mget k1 k4 k7, mget k2 k5 k8 & mget k3 k6 k9. This is significantly better than making one message per key and the performance advantage gets better with more keys in the query.

Quorum:
Of course, things get complicated with quorum. For example, a subquery of the previous query like this mget k1 k4 k7, can yield results from 3 replicas like v1 v4 nil, v1 nil v7, nil v4 v7. The checksums of the three responses will not match in which case, Dynomite will split the values and try to get a quorum on individual values (ex. v1, v1 and nil) and then assemble the final response for the subquery (mget k1 k4 k7) making it v1 v4 v7.

Error Scenarios:
In this pull request, Dynomite fails the entire request if one of the nodes fail. This is because Jedis fails to cast an intermediate error and throws an exception.
For example in case of a response like this:

127.0.0.1:8102> mget GetSet_140_81 GetSet_140_28 GetSet_140_71 GetSet_140_972 GetSet_140_87 GetSet_140_777
1) "GetSet_140_81__value__GetSet_140_81"
2) (error) ERR Peer: Peer Node is not connected
3) "GetSet_140_71__value__GetSet_140_71"
4) "GetSet_140_972__value__GetSet_140_972"
5) (error) ERR Peer: Peer Node is not connected
6) "GetSet_140_777__value__GetSet_140_777"
127.0.0.1:8102>

Jedis cannot handle this but redis-cli can. Since jedis is the most widely used client, I decided to fail the entire request instead like this:

127.0.0.1:8102> mget GetSet_140_81 GetSet_140_28 GetSet_140_71 GetSet_140_972 GetSet_140_87 GetSet_140_777
(error) ERR Peer: Peer Node is not connected

### DIFF
--- a/src/dyn_dnode_msg.c
+++ b/src/dyn_dnode_msg.c
@@ -410,7 +410,7 @@ dyn_parse_req(struct msg *r)
 {
 	if (log_loggable(LOG_VVERB)) {
 		log_debug(LOG_VVERB, ":::::::::::::::::::::: In dyn_parse_req, start to process request :::::::::::::::::::::: ");
-		msg_dump(r);
+		msg_dump(LOG_VVERB, r);
 	}
 
 	bool done_parsing = false;
@@ -501,7 +501,7 @@ dyn_parse_req(struct msg *r)
 	//bad case
 	if (log_loggable(LOG_VVERB)) {
 		log_debug(LOG_VVERB, "Bad or splitted message");  //fix me to do something
-		msg_dump(r);
+		msg_dump(LOG_VVERB, r);
 	}
 	r->result = MSG_PARSE_AGAIN;
 }
@@ -511,7 +511,7 @@ void dyn_parse_rsp(struct msg *r)
 {
 	if (log_loggable(LOG_VVERB)) {
 		log_debug(LOG_VVERB, ":::::::::::::::::::::: In dyn_parse_rsp, start to process response :::::::::::::::::::::::: ");
-		msg_dump(r);
+		msg_dump(LOG_VVERB, r);
 	}
 
 	bool done_parsing = false;
@@ -588,7 +588,7 @@ void dyn_parse_rsp(struct msg *r)
 	//bad case
 	if (log_loggable(LOG_DEBUG)) {
 		log_debug(LOG_DEBUG, "Resp: bad message - cannot parse");  //fix me to do something
-		msg_dump(r);
+		msg_dump(LOG_DEBUG, r);
 	}
 
 	r->result = MSG_PARSE_AGAIN;
--- a/src/dyn_dnode_msg.c
+++ b/src/dyn_dnode_msg.c
@@ -279,6 +279,7 @@ dyn_parse_core(struct msg *r)
 
       case DYN_DONE:
          log_debug(LOG_VVERB, "DYN_DONE");
+         r->mlen -= (p - r->pos);
          r->pos = p;
          dmsg->payload = p;
          r->dyn_parse_state = DYN_DONE;
--- a/src/dyn_dnode_msg.c
+++ b/src/dyn_dnode_msg.c
@@ -279,7 +279,7 @@ dyn_parse_core(struct msg *r)
 
       case DYN_DONE:
          log_debug(LOG_VVERB, "DYN_DONE");
-         r->mlen -= (p - r->pos);
+         r->mlen -= (uint32_t)(p - r->pos);
          r->pos = p;
          dmsg->payload = p;
          r->dyn_parse_state = DYN_DONE;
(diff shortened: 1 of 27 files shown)


src/proto/dyn_memcache.c
```diff
@@ -346,29 +348,60 @@ memcache_parse_req(struct msg *r)
...
 
         case SW_KEY:
+            if (r->token == NULL) {
+                r->token = p;
+            }
             if (ch == ' ' || ch == CR) {
-                if ((p - r->key_start) > MEMCACHE_MAX_KEY_LENGTH) {
+                struct keypos *kpos;
+                int keylen = p - r->token;
+                if (keylen > MEMCACHE_MAX_KEY_LENGTH) {
```
ipapapa, line 364: How are the memcached protocol changes related to the multikey operations? Have these updates been tested? Do we have a test case that covers the memcached implementation? If yes to any of these questions, then can we resolved https://github.com/Netflix/dynomite/issues/220 ?

src/proto/dyn_memcache.c
```diff
@@ -763,17 +792,17 @@ memcache_parse_rsp(struct msg *r)
         SW_RSP_STR,
         SW_SPACES_BEFORE_KEY,
         SW_KEY,
-        SW_SPACES_BEFORE_FLAGS,
+        SW_SPACES_BEFORE_FLAGS,     /* 5 */
```
ipapapa, line 795: What are the numbers in the comments?

src/proto/dyn_memcache.c
shailesh33: I picked the memcache changes as is from the twemproxy code.

src/proto/dyn_memcache.c
shailesh33: I picked the memcache changes as is from the twemproxy code.

src/dyn_dnode_msg.c
```diff
@@ -279,6 +279,7 @@ dyn_parse_core(struct msg *r)
 
       case DYN_DONE:
          log_debug(LOG_VVERB, "DYN_DONE");
+         r->mlen -= (p - r->pos);
```
ipapapa: verify?

src/dyn_mbuf.c
```diff
@@ -144,6 +144,18 @@ mbuf_free_queue_size(void)
 }
 
 
+void mbuf_dump_pos(struct mbuf *mbuf)
```
ipapapa: probably consider clubbing them together?

src/dyn_request.c
```diff
@@ -95,8 +95,11 @@ req_done(struct conn *conn, struct msg *req)
         return true;
     }
 
-    /* check all fragments of the given request vector are done */
+    if (req->nfrag_done < req->nfrag)
+        return false;
 
+    /* check all fragments of the given request vector are done */
```
ipapapa: mixing comment style.

src/proto/dyn_redis.c
```diff
@@ -2591,14 +2582,113 @@ redis_pre_coalesce(struct msg *rsp)
...
+    }
+
+    for (i = 0; i < array_n(request->keys); i++) {      /* for each key */
+        sub_msg = request->frag_seq[i]->selected_rsp;           /* get it's peer response */
+        if (sub_msg == NULL) {
+            log_warn("marking %M as error", response->owner);
+            response->owner->err = 1;
+            return;
+        }
+        //TODO: check if the selected rsp is an error or what?
```
ipapapa: TODO?

src/proto/dyn_redis.c
```diff
@@ -2591,14 +2582,113 @@ redis_pre_coalesce(struct msg *rsp)
...
+    case MSG_REQ_REDIS_MSET:
+        return redis_post_coalesce_mset(req);
+
+    default:
+        NOT_REACHED();
+    }
+}
+
+
+void
```
ipapapa: Remove this code if not used.

src/proto/dyn_redis.c
```diff
@@ -2649,3 +2739,243 @@ redis_post_coalesce(struct msg *req)
...
+redis_fragment(struct msg *r, struct server_pool *pool, struct rack *rack, struct msg_tqh *frag_msgq)
+{
+    if (1 == array_n(r->keys)){
+        return DN_OK;
+    }
+
+    switch (r->type) {
+    case MSG_REQ_REDIS_MGET:
+    case MSG_REQ_REDIS_DEL:
+        return redis_fragment_argx(r, pool, rack, frag_msgq, 1);
```
ipapapa, line 2928: Replace 1 with a defined value or add a comment

ipapapa (approved):

src/dyn_dnode_msg.c
shailesh33: so we reach this stage when we have parsed the Dynomite header, this change is to reduce the length of the message by the header size. We didn't see this issue before because this `mlen` was not used previously. When we copy results from different nodes to consolidate a response for `MGET`, we started using this field.

src/dyn_mbuf.c
shailesh33: I have already clubbed them

src/dyn_request.c
shailesh33: done

src/proto/dyn_redis.c
shailesh33: done.

src/proto/dyn_redis.c
shailesh33: done

src/proto/dyn_redis.c
shailesh33: will pass on this

PR #483, opened 2017-09-21
Fix memory corruption with peer array

shailesh33:
We create a peers array in the server pool which was an array of struct
node. The address of this struct was being handed all over which can
become invalid when the array resizes and reallocates the underlying
memory. The solution is to have arrays as an array of pointers and hand
over those pointers.

There is a need for a special type of array which will hold only
pointers. That will simplify a lot of code

### DIFF
--- a/src/dyn_dnode_peer.c
+++ b/src/dyn_dnode_peer.c
@@ -173,10 +173,7 @@ dnode_peer_deinit(struct array *nodes)
     uint32_t i, nnode;
 
     for (i = 0, nnode = array_n(nodes); i < nnode; i++) {
-        struct node *s;
-
-        s = array_pop(nodes);
-        IGNORE_RET_VAL(s);
+        struct node *s = *(struct node **)array_pop(nodes);
         if (s->conn_pool) {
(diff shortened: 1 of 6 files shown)


src/dyn_dnode_peer.c
```diff
@@ -268,11 +269,14 @@ dnode_initialize_peers(struct context *ctx)
...
 
-    THROW_STATUS(array_init(peers, nseed + 1, sizeof(struct node)));
+    THROW_STATUS(array_init(peers, nseed + 1, sizeof(struct node *)));
 
     // Add self node
-    struct node *self = array_push(peers);
-    ASSERT(self != NULL);
+    struct node **selfptr = array_push(peers);
+    struct node *self = dn_zalloc(sizeof(struct node));
+    if (!self || !selfptr)
```
ipapapa, line 277: why not use `ASSERT` over returning with `DN_ENOMEM`?

shailesh33: so that you can do the clean in the caller function.

ipapapa (approved):

PR #489, opened 2017-10-12
Fix crash when decrypting large payload data

shailesh33:
This was occurring frequently in a production environment where large payload in the range of 40K was caused Dynomite to crash.

### DIFF
--- a/src/dyn_message.c
+++ b/src/dyn_message.c
@@ -166,8 +166,8 @@ int
 print_req(FILE *stream, struct msg *req)
 {
     struct string *req_type = msg_type_string(req->type);
-    return fprintf(stream, "<REQ %p %lu:%lu %.*s>", req, req->id, req->parent_id,
-                   req_type->len, req_type->data);
+    return fprintf(stream, "<REQ %p %lu:%lu::%lu %.*s len:%u>", req, req->id, req->parent_id,
+                   req->frag_id, req_type->len, req_type->data, req->mlen);
 }
 
 int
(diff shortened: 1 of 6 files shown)


src/dyn_message.c
```diff
@@ -986,8 +986,37 @@ msg_recv_chain(struct context *ctx, struct conn *conn, struct msg *msg)
          msg->dmsg->flags & 0x1) ? msg->dmsg->plen : -1;  //used in encryption case only
 
     mbuf = STAILQ_LAST(&msg->mhdr, mbuf, next);
-    if (mbuf == NULL || mbuf_full(mbuf) ||
-        (expected_fill != -1 && mbuf->last == mbuf->end_extra)) {
+    /* This logic is unncessarily complicated. Ideally a connection should read
```
ipapapa, line 989: I do not understand this comment. The logic is wrong. If you start decrypting before you have received the full buffer then you will not be able to properly decrypt

src/dyn_message.c
```diff
@@ -1035,7 +1064,7 @@ msg_recv_chain(struct context *ctx, struct conn *conn, struct msg *msg)
                 }
 
                 status = dyn_aes_decrypt(mbuf->start, mbuf->last - mbuf->start, nbuf, msg->owner->aes_key);
-                if (status == DN_OK) {
+                if (status >= DN_OK) {
```
ipapapa, line 1067: I think there is some overloading on what `rstatus_t` is supposed to be. What does `>=DN_OK` mean here? I understand that `dyn_aes_decrypt` is returning the length and therefore fundamentally it should not return an `rstatus_t` which is a typedef.

src/dyn_message.c
shailesh33: We move from a buffer to buffer. For a message of size say 40K, and a buffer of size 16K, while sending we encrypt the first 16k and send, then next 16K and send, then final 8k.
While decrypting, we receive the first 16K, decrypt, then next 16K then decrypt and then 8k and decrypt. This is faster because we will decrypt while we are waiting for the next buffer. But it complicates the matters.
Instead, it will be better if we wait for 40K and then decrypt it all. 
But now I understand that it is in the favor of how the underlying Redis parser works.

src/dyn_message.c
shailesh33: yes, you got it correct. we return <0 if there is an error and >0 if success. So THis check was pretty much wrong. I seriously do not know if we are covering all the different error scenarios here.

ipapapa (approved):

PR #490, opened 2017-10-25
reduce compiler warnings

shailesh33:
no major changes just added some typecasts etc wherever necessary
tried to change some variable types where appropriate
reduced some logging levels that I accidentally bumped up
Simplified some logic in receiving encrypted data
Took care of some shadow declarations.

Overall the new code is fairly warning free. The old code still has some warnings


src/dyn_core.c
```diff
@@ -64,7 +64,6 @@ core_print_peer_status(void *arg1)
             }
         }
     }
```
ipapapa, line 66: why don't we need this task anymore?

src/dyn_core.h
```diff
@@ -294,7 +294,7 @@ struct server_pool {
     int                backlog;              /* listen backlog */
     uint32_t           client_connections;   /* maximum # client connection */
     msec_t             server_retry_timeout_ms; /* server retry timeout in msec */
-    uint32_t           server_failure_limit; /* server failure limit */
+    uint8_t            server_failure_limit; /* server failure limit */
```
ipapapa, line 297: up to 255 only?

src/proto/dyn_redis.c
```diff
@@ -2579,6 +2579,7 @@ redis_pre_coalesce(struct msg *rsp)
          * MSG_RSP_REDIS_MULTIBULK. For an invalid response, we send out -ERR
          * with EINVAL errno
          */
+        log_info("Invalid Response type");
```
ipapapa: should this be a warning?

ipapapa (approved):

src/dyn_core.c
shailesh33: too much logging. every 30 secs

src/dyn_core.h
shailesh33: not enough?

PR #494, opened 2017-11-08
Use a cluster generator to automatically create and test clusters

hashbrowncipher:
To accommodate a desire to be able for developers to easily test against a variety of Dynamo rings, this change adds a Python script which autogenerates clusters, launches them, and runs the existing functional tests against them.

One goal that I kept in mind is to generate clusters deterministically based on the request file. Right now the generation is fully deterministic, but I can also make it non-deterministic, but print the PRNG state to stdout. This would allow a developer to reproduce a previous test run by passing the desired PRNG state as a shell argument.

Potential follow-ups:
1. Many files in the conf directory may now be obsolete, but I haven't deleted them as I wasn't sure if they had other uses.
1. Right now the functional testing script doesn't know about differences in cluster topology. Since it's Python, perhaps the easiest thing to do would be to import the script as a Python module rather than running it as an executable.
1. If we upgrade to Python 3, contextlib.ExitStack would make it easier to ensure that all subprocesses are killed when the process exits.


ipapapa:
* I suggest we delete the unused YAML files if not used. @shailesh33?
* @hashbrowncipher Why don't we upgrade to Python 3 then?

shailesh33:
Python is mainly used for testing here so if you think upgrading is the right thing, go for it.

test/cluster_generator.py
```diff
@@ -0,0 +1,211 @@
...
+        # filter out our own seed string
+        conf['dyn_seeds'] = filter(lambda s: s != self.seed_string, seeds_list)
+        conf['servers'] = ['{}:{}:0'.format(self.ip, REDIS_PORT)]
+        conf['stats_listen'] = '{}:{}'.format(self.ip, STATS_PORT)
+        conf['tokens'] = self.token
+        conf['local_peer_connections'] = self.local_connections
+        conf['remote_peer_connections'] = self.remote_connections
+        return dict(dyn_o_mite=conf)
+
+    def write_config(self, seeds_list):
```
ipapapa, line 119: We have another script for people to generate YAML files: https://github.com/Netflix/dynomite/blob/dev/scripts/dynomite/generate_yamls.py Shall we combine them together, or at least somehow leverage each other to avoid duplicate scripts?

hashbrowncipher: Will do.

hashbrowncipher:
I'm making the PR to upgrade to Py3 right now.

test/func_test.py
```diff
@@ -134,12 +134,12 @@ def main(args):
     # This test assumes for now that the nodes are running at the given ports.
     # This is done by travis.sh. Please check that file and the corresponding
     # yml files for each dynomite instance there to get an idea of the topology.
-    r = RedisNode(host="localhost", ip="127.0.0.1", port=1212)
-    d1 = DynoNode(host="127.0.0.1", ip="127.0.0.1", data_store_port=22121)
-    d2 = DynoNode(host="127.0.0.2", ip="127.0.0.2", data_store_port=22122)
-    d3 = DynoNode(host="127.0.0.3", ip="127.0.0.3", data_store_port=22123)
-    d4 = DynoNode(host="127.0.0.4", ip="127.0.0.4", data_store_port=22124)
-    d5 = DynoNode(host="127.0.0.5", ip="127.0.0.5", data_store_port=22125)
+    r = RedisNode(host="127.0.1.1", ip="127.0.1.1", port=1212)
```
shailesh33: Whatever we got till now is great. However, this whole section is still hardcoded. Can it understand what dynamic cluster is created by cluster generator? Ideally, I would expect DynoNode to have a start/launch api that will start that particular node. 
DynoCluster consists of many DynoNodes which have start/stop APIs etc.

hashbrowncipher: I believe I've remedied this in #495. Let me know if I've achieved the desired design.

test/cluster_generator.py
shailesh33: I guess you will do this in a separate PR?

PR #495, opened 2017-11-09
Upgrade to Python 3 + plumb cluster_generator into func_test

hashbrowncipher:
This PR is based on the branch in #494. It achieves the second and third follow-ups from #494's PR message.

### Follow-up 2: Plumbing

I hooked up the cluster_generator module into func_test by exposing a new method in func_test which allows for a RedisNode and a DynoCluster to be passed in, and runs tests with the connection information from those objects.

### Follow-up 3: Python 3 + ExitStack

I ran 2to3 (Python 2 -> 3 conversion tool) and then fixed a few straggler issues that it didn't (one was strings.letters -> ascii_letters; another was a change in how Python handles division).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,2 +1,8 @@
 language: c
 script: bash ./travis.sh
+
+addons:
+  apt:
+   packages:
+   - python3
+   - python3-pip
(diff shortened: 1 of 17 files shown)


ipapapa (approved):

shailesh33 (approved):
Looks good to me

PR #498, opened 2017-11-15
EVAL support in Dynomite

shailesh33:
Parse the EVAL request correctly. Also, restrict the keys in the script to belong to
same node.

### DIFF
--- a/src/dyn_message.h
+++ b/src/dyn_message.h
@@ -37,6 +37,8 @@
 typedef void (*func_msg_parse_t)(struct msg *, const struct string *hash_tag);
 typedef rstatus_t (*func_msg_fragment_t)(struct msg *, struct server_pool *,
                                          struct rack *, struct msg_tqh *);
+typedef rstatus_t (*func_msg_verify_t)(struct msg *, struct server_pool *,
+                                         struct rack *);
 typedef void (*func_msg_coalesce_t)(struct msg *r);
 typedef rstatus_t (*msg_response_handler_t)(struct msg *req, struct msg *rsp);
 typedef bool (*func_msg_failure_t)(struct msg *r);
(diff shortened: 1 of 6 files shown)


src/dyn_message.h
```diff
@@ -261,6 +265,8 @@ dn_strerror(dyn_error_t err)
             return "Datastore refused connection";
         case DYNOMITE_NO_QUORUM_ACHIEVED:
             return "Failed to achieve Quorum";
+        case DYNOMITE_SCRIPT_SPANS_NODES:
+            return "Script Spans multiple Dynomite nodes";
```
ipapapa: Usually, a failure message contains some negativity like "refused" or "failed". In this error message, I am not sure the client user will understand what it means.

shailesh33: good catch!. I will edit it as ```Keys in the script cannot span multiple Dynomite nodes```

src/proto/dyn_redis.c
```diff
@@ -2938,6 +2951,31 @@ redis_fragment(struct msg *r, struct server_pool *pool, struct rack *rack, struc
     }
 }
 
+rstatus_t
+redis_verify_request(struct msg *r, struct server_pool *pool, struct rack *rack)
```
ipapapa, line 2955: `redis_verify_request` is very generic. Should it contain something about this being about LUA. At least a comment or some indication for the maintainer.

shailesh33: There is a comment in the code. Also this is a general function to verify every request, some are handled, some are now. Just like redis_fragment,

src/dyn_message.h
shailesh33: Fixed it. Now it looks like this:

```
127.0.0.1:8102> eval "return {redis.call('get',KEYS[1]),ARGV[1],ARGV[2]}" 3 key:000000058729 key:000000091650 key:000000019267 arg1 arg2
(error) ERR Dynomite: Keys in the script cannot span multiple nodes
```

PR #515, opened 2017-12-16
Removed printf custom specifier (revert back musl compatibility)

UladzimirTrehubenka:
Back port logging from dev to v0.5.9


shailesh33:
This commit on Dev broke the build on APPLE. Can you also merge the last commit I did in dev to fix that Apple build?

UladzimirTrehubenka:
Apple build fix from dev branch was added.

PR #516, opened 2017-12-16
Add DYNOMITE_DNS_NAME and DYNOMITE_DNS_TYPE env vars for dns_provider

UladzimirTrehubenka:
Behaviour:
1) no env var:
T_TXT "_dynomite.ec2-internal" (default value)
2) export DYNOMITE_DNS_NAME="_dynomite.internal"
T_TXT "_dynomite.internal" (override name)
3) export DYNOMITE_DNS_NAME="k8s.dynomite.internal"
export DYNOMITE_DNS_TYPE="A"
T_A "k8s.dynomite.internal" (override name and use A instead TXT)

NOTE: DYNOMITE_DNS_TXT_NAME was renamed to DYNOMITE_DNS_NAME
For DYNOMITE_DNS_TYPE only "A" value is supported,
another value is ignored (query type remains T_TXT)

### DIFF
--- a/src/seedsprovider/dyn_dns.c
+++ b/src/seedsprovider/dyn_dns.c
@@ -27,7 +27,9 @@
 #define DNS_TXT_NAME "_dynomite.ec2-internal"
 #endif
 
-static char * txtName  = NULL;
+static char * dnsName  = NULL;
+static char * dnsType  = NULL;
+static int queryType = T_TXT;
 static int64_t last = 0; //storing last time for seeds check
 static uint32_t last_seeds_hash = 0;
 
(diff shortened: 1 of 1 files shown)


shailesh33:
The changes look fine. I am wondering what should be the behavior when both environment variables are set. Should we error out so the behavior is crisply defined?

UladzimirTrehubenka:
Probably better way is to use two env vars - one for record type (DYNOMITE_DNS_TYPE - default T_TXT) and another one for name (DYNOMITE_DNS_NAME - default _dynomite.ec2-internal).

shailesh33:
I like that approach. Can you please make the changes. Make sure it is backward compatible.

UladzimirTrehubenka:
BTW for full backward compatibility need to use existing DYNOMITE_DNS_TXT_NAME and new DYNOMITE_DNS_TYPE but DYNOMITE_DNS_TXT_NAME is confusing here.

shailesh33:
apologies for the delay.
the changes look good to me.

robertofabrizi:
I've set the following in my dyno config:

`  dyn_seed_provider: dns_provider
`
and this in the ENV:

```
      - DYNOMITE_DNS_NAME=dynomite-service
      - DYNOMITE_DNS_TYPE=A
```

but I receive this error: [2018-06-08 11:20:38.976] dns_get_seeds:124 invalid length for dynomite-service: 10 < 4

The line of code that triggers it is mysterious to me...

```
unsigned char *s = ns_rr_rdata(rr);
        if (s[0] >= ns_rr_rdlen(rr)) {
            log_debug(LOG_DEBUG, "invalid length for %s: %d < %d", dnsName, s[0], ns_rr_rdlen(rr));
            return DN_NOOPS;
        }
```

Can anyone help me understand the issue?
EDIT: dynomite-service is resolved correctly to 10.0.2.4

Thanks everyone!

UladzimirTrehubenka:
Hm, need to split handling A RR and TXT RR.
A record RDATA format is:
```
ip1, 4 byte
ip2, 4 byte
etc.
```
TXT record RDATA format is:
```
length, 1 byte
string, <length bytes>
length2, 1 byte
string2, <length2 bytes>
etc.
```
In this case ns_rr_rdlen(rr)=4 means that A RR contains one IP address (4 byte).
s[0]=10 means first octet in IP address (s[1]=0, s[2]=2, s[3]=4)

robertofabrizi:
@UladzimirTrehubenka  First of all thank you so much for your time.
At the risk of sounding like a fool, which I probably am, I don't really understand the point of that if statement.
It checks that the first octet of the ip 10.0.2.4 is bigger than the number of bytes in one ip? What's the point of that? Also, how can it be working for everyone except maybe those whose ip returned by the dns service doesn't start with a 2.x.x.x ?

EDIT_: ah ok maybe I got your point, you are saying that this check doesn't make sense for A records. Should I just remove the if from the source before rebuilding it?

UladzimirTrehubenka:
Line ```if (s[0] >= ns_rr_rdlen(rr))``` is needed for TXT only, so need to add type check (A or TXT).
For make hash line ```mbuf_copy(seeds_buf, s + 1, s[0]);``` is valid for TXT only, so for A RR need to use something like this: ```mbuf_copy(seeds_buf, ns_rr_rdata(rr), ns_rr_rdlen(rr));``` (BTW usually resolver returns IPs in random order - so need to sort IPs first to avoid false trigger that seeds list was changed).
Also ```log_debug(LOG_VERB, "seeds for %s: %.*s", dnsName, s[0], s +1);``` is valid for TXT only,
so need to transform IP to human readable string (BTW A RR can contain more than one IP).

robertofabrizi:
I've applied the fix and now I have no errors in the logs, yet when I ask dynomite the data and the kubernetes routes me to container 1 or 2 in a round robin fashion, only when I land on the container that actually received the set I get the value with the get, when I end up on the other container I get null....

UladzimirTrehubenka:
Could you please raise an issue and create PR with your fix?

PR #522, opened 2018-01-15
Fix crashes caused by parsing unexpected Florida responses

sobkovych:


sobkovych:
@shailesh33 
Could you please help to resolve this weird CI error?
I have no idea how to fix it.
Thanks!

shailesh33:
some thing was wrong with travis earlier. Seems to be working now.

src/dyn_gossip.c
```diff
@@ -698,21 +704,23 @@ gossip_update_seeds(struct server_pool *sp, struct mbuf *seeds)
...
 
-    while (q > start) {
+    while (q != NULL && q > start) {
         seed_node = q + 1;
         seed_node_len = (uint32_t)(p - seed_node + 1);
         string_copy(&temp, seed_node, seed_node_len);
         //array_init(&tokens, 1, sizeof(struct dyn_token));
         init_dyn_token(&token);
-        parse_seeds(&temp, &dc_name, &rack_name, &port_str, &address, &ip,  &token);
+        parse_status = parse_seeds(&temp, &dc_name, &rack_name, &port_str, &address, &ip,  &token);
```
shailesh33: what happens when parse_status is GOS_ERROR, what gets printed below, and do we proceed to the next seed?

sobkovych: On GOS_ERROR will be printed values for successfully parsed fields or empty strings otherwise.
We will continue to proceed next seed(s) up to all '|' delimiters within a parsed string.
I've added one more commit to fix some other crashes related to Florida responses parsing and align printed logs.

shailesh33 (approved):

PR #525, opened 2018-01-24
Improve florida html parser

sobkovych:
These changes help to deal with a case when HTML response might be unexpectedly splitted into parts directly on HTML content delimiter ("\r\n\r\n" sequence). The general idea here is below:
* before every parsing iteration read socket until getting all data or filling RX buffer
* if we don't get all data at current iteration and HTML content beginning still wasn't detected, last 3 received bytes will be copied to the beginning of the RX buffer and will be used on the next parsing iteration

### DIFF
--- a/src/seedsprovider/dyn_florida.c
+++ b/src/seedsprovider/dyn_florida.c
@@ -137,25 +137,27 @@ florida_get_seeds(struct context * ctx, struct mbuf *seeds_buf) {
     uint8_t * htmlcontent;
     uint8_t *ok = NULL;
 
-    //assume that the respsone payload is under BUF_SIZE
-    while ((tmpres = recv(sock, buf, BUFSIZ, 0)) > 0) {
+    bool socket_has_data = true;
+    uint32_t rx_total = 0;
 
-        // Look for a OK response  in the first buffer output.
+    while (socket_has_data) {
+        // Read socket data until we get them all or RX buffer becomes full
+        while ((rx_total < BUFSIZ) && (tmpres = recv(sock, buf + rx_total, BUFSIZ - rx_total, 0)) > 0) {
+            rx_total += tmpres;
+        }
+
+        // Look for a OK response in the first buffer output.
         if (!ok)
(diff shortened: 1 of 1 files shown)


src/seedsprovider/dyn_florida.c
```diff
@@ -165,11 +167,27 @@ florida_get_seeds(struct context * ctx, struct mbuf *seeds_buf) {
...
+        if (htmlstart) {
+            mbuf_copy(seeds_buf, htmlcontent, rx_total - (htmlcontent - buf));
         }
 
-        memset(buf, 0, tmpres);
+        // If socket still has data for reading
+        if (tmpres > 0) {
+            if ((htmlstart == 0) && (rx_total >= 3)) {
+                /* Under certain conditions \r\n\r\n part might by splitted into two
+                 * messages, so copy last 3 received bytes to the buf start to be able
```
shailesh33: I am not exactly sure of why `3`. Can you please elaborate?

sobkovych: I know this part looks a bit weird, but in some corner cases reproduced in my tests (when read buffer size was close to response header size) I've observed situations when '\r\n\r\n' sequence was splitted and because of that, we couldn't detect html content start. To deal with this case we need to move only last 3 bytes from current read iteration to buffer head before the next read iteration, so this will save us from losing our sequence. With this changes, response parsing works fine on any responses size and with any read buffer size (>= 4 bytes). Please notice, we do this only before html content was found.

sobkovych:
@shailesh33 
Please, take a look at the [visual explanation](https://drive.google.com/drive/folders/11r1tpPU1D2NycGgOwcypoorl1jH08rV2?usp=sharing) why we should copy at least 3 bytes between split packet parts.

src/seedsprovider/dyn_florida.c
shailesh33: can you please add this to the comment in the code? I guess then we are done.

shailesh33 (approved):

sobkovych:
@shailesh33 
Looks like some tests failed. Any idea how to fix things?

PR #536, opened 2018-02-19
Treat generated AES key as an array

sobkovych:
Due to a possibility of '\0' inside generated AES key we can't use "str-family" functions on key processing (e.g. strlen, strncpy etc.) and should treat key as an array. Otherwise, in some cases we get an error in dyn_aes_decrypt() routine ("Bad data in EVP_DecryptFinal_ex, crypto data with XX bytes of data").

### DIFF
--- a/src/dyn_connection_internal.c
+++ b/src/dyn_connection_internal.c
@@ -143,8 +143,12 @@ _conn_get(void)
     conn_set_write_consistency(conn, g_write_consistency);
     conn->type = CONN_UNSPECIFIED;
 
+    // Generate a new key for each connection
     unsigned char *aes_key = generate_aes_key();
-    strncpy((char *)conn->aes_key, (char *)aes_key, strlen((char *)aes_key)); //generate a new key for each connection
+    if (aes_key == NULL) {
+        return NULL;
+    }
+    memcpy(conn->aes_key, aes_key, AES_KEYLEN);
 
     return conn;
 }
--- a/src/dyn_connection_internal.c
+++ b/src/dyn_connection_internal.c
@@ -146,6 +146,7 @@ _conn_get(void)
     // Generate a new key for each connection
     unsigned char *aes_key = generate_aes_key();
     if (aes_key == NULL) {
+        dn_free(conn);
         return NULL;
     }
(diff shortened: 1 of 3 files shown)


sobkovych:
@shailesh33 
Hi, is there any chance to review this? Should the pull request be assigned to someone when creating?
Thanks!

src/dyn_connection_internal.c
```diff
@@ -143,8 +143,12 @@ _conn_get(void)
     conn_set_write_consistency(conn, g_write_consistency);
     conn->type = CONN_UNSPECIFIED;
 
+    // Generate a new key for each connection
     unsigned char *aes_key = generate_aes_key();
-    strncpy((char *)conn->aes_key, (char *)aes_key, strlen((char *)aes_key)); //generate a new key for each connection
+    if (aes_key == NULL) {
+        return NULL;
```
shailesh33: you should free conn here

shailesh33 (changes requested):
Please fix that memory leak. Rest looks good to me

sobkovych:
@shailesh33 
Thanks, great! Fixed.

sobkovych:
@shailesh33 
Hi, is there any chance to finish with this PR?

src/dyn_connection_internal.c
```diff
@@ -143,8 +143,13 @@ _conn_get(void)
     conn_set_write_consistency(conn, g_write_consistency);
     conn->type = CONN_UNSPECIFIED;
 
+    // Generate a new key for each connection
     unsigned char *aes_key = generate_aes_key();
-    strncpy((char *)conn->aes_key, (char *)aes_key, strlen((char *)aes_key)); //generate a new key for each connection
+    if (aes_key == NULL) {
+        dn_free(conn);
```
shailesh33: I am not exactly sure but may be you should use _conn_put() here. Infact may be generate the aes key at the beginning of the function and fail right there.

shailesh33 (approved):
it will be great if you can do those changes. I am approving nevertheless.

sobkovych:
Yes, you're absolutely right! Done, please take a look.
Thanks!

shailesh33 (approved):

PR #558, opened 2018-06-05
OpenSSL 1.1 compatibility

i-rinat:
OpenSSL 1.1 makes EVP_CIPHER_CTX an opaque struct, so a special function call is needed to allocate it. Library initialization was simplified a bit.

### DIFF
--- a/src/dyn_crypto.c
+++ b/src/dyn_crypto.c
@@ -126,15 +126,23 @@ static rstatus_t
 aes_init(void)
 {
     // Initialize contexts
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
     aes_encrypt_ctx = (EVP_CIPHER_CTX*) malloc(sizeof(EVP_CIPHER_CTX));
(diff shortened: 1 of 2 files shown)


i-rinat:
Related OpenSSL documentation pages:
* https://www.openssl.org/docs/man1.1.0/crypto/OpenSSL_add_all_algorithms.html
* https://www.openssl.org/docs/man1.1.0/crypto/EVP_EncryptInit.html

i-rinat:
(Github interface said that the branch is not up-to-date, so I rebased it on top of the current `dev` state. No other changes were made.)

i-rinat:
ping

shailesh33:
Looks good to me.

shailesh33:
@i-rinat I assume the two versions interoperate. In the sense that one node with older version can talk to a node with a newer version and vice versa,

i-rinat:
> In the sense that one node with older version can talk to a node with a newer version and vice versa,

I didn't check that, but I guess there should be no differences in the actual algorithms. Otherwise newer versions of cURL would fail to talk to older HTTPS severs, for example.

PR #590, opened 2018-08-20
Avoid accessing 'ctx->stats' if it was never created

smukil:
If a call to 'stats_create()' fails from 'core_stats_create()', we
still access the 'stats' structure on the teardown path in multiple
places. This causes Dynomite to crash in multiple places.

This patch fixes the crashes by avoiding accessing the invalid memory
if it wasn't allocated in the first place.

### DIFF
--- a/src/dyn_core.c
+++ b/src/dyn_core.c
@@ -229,8 +229,9 @@ core_crypto_init(struct context *ctx)
     /* crypto init */
     THROW_STATUS(crypto_init(&ctx->pool));
     rstatus_t status = core_stats_create(ctx);
-    if (status != DN_OK)
-	stats_destroy(ctx->stats);
+    if (status != DN_OK) {
+        if (ctx->stats) stats_destroy(ctx->stats);
+    }
     
     return status;
 }
@@ -267,6 +268,7 @@ core_ctx_create(struct instance *nci)
         loga("Failed to create context!!!");
         return DN_ERROR;
     }
+
     nci->ctx = ctx;
     ctx->instance = nci;
     ctx->cf = NULL;
--- a/src/dyn_server.c
+++ b/src/dyn_server.c
@@ -144,8 +144,10 @@ static void
 server_failure(struct context *ctx, struct datastore *server)
 {
     conn_pool_notify_conn_errored(server->conn_pool);
-    stats_server_set_ts(ctx, server_ejected_at, dn_msec_now());
-    stats_pool_incr(ctx, server_ejects);
+    if (ctx->stats) {
+        stats_server_set_ts(ctx, server_ejected_at, dn_msec_now());
+        stats_pool_incr(ctx, server_ejects);
+    }
 }
 
 static void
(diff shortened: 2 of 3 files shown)


src/dyn_core.c
```diff
@@ -229,8 +229,9 @@ core_crypto_init(struct context *ctx)
     /* crypto init */
     THROW_STATUS(crypto_init(&ctx->pool));
     rstatus_t status = core_stats_create(ctx);
-    if (status != DN_OK)
-	stats_destroy(ctx->stats);
+    if (status != DN_OK) {
```
shailesh33, line 232: why not check stats in the stats_* functions?

smukil: So, I thought about that approach as well and this was my thinking. Let me know if you agree or not:

There should be a contract between the 'stats_\*' class of functions and their users (which in this case is dyn_core and dyn_server).

The contract should specify one of the following 2 options:
1) The 'stats_\*' functions may assume that the 'struct stats' is always valid and exists, and so if any 'stats_\*' function is called, that means that the 'struct stats' has successfully been setup by the caller of these functions, and that means that the caller guarantees the validity of the struct. I took this approach and this is why I added the ASSERT() in L1790 in dyn_stats.c.

2) The alternative is, the callers of the 'stats_\*' functions do not need to assume anything about the validity of the 'struct stats' object , and the 'stats_\*' functions themselves make sure not to take any action if the 'struct stats' object is NULL. The reason I didn't go for this is because it seemed that all the 'stats_\*' functions were already assuming that the 'struct stats' would be valid and didn't want to add an "if (ctx->stats != NULL)" to the beginning of every 'stats_\*' function.

Also, over time, I feel like we can make contracts like the ones above more explicit so as to avoid accidentally introducing bugs like this.

shailesh33: looks like a valid approach.

ipapapa (approved):

src/dyn_core.c
ipapapa: I also like the approach of specifying a contract.  I also had the same question as the caller code may get overloaded checking for `ctx->stats!=NULL` for every `stats_*` call. Actually if we follow this path for every called function, caller codes would become overloaded. The other way around is to have every `stats_*` do the check. I am not sure which one is better.

smukil: @ipapapa Yes that's true. I haven't gone too deep into the code yet, but ideally the way the code would be setup is that the teardown path should never have any failures possible.
Eg: malloc() can return NULL to indicate it failed, but 'void free()' should never ever fail if we're obeying its contract. The contract for free() is that if you pass it an invalid value, it will crash the process, and that is well documented. This is obviously an oversimplified example and in our case the code is more complicated, but we can still try to follow the same principles as much as the code will allow us.

Similarly, we should be able to refactor parts of the code such that the teardown paths have well documented contracts, so that once we do such a refactor, we won't have to add these if() else() checks in too many places.

I'll have to understand the code a little better to be able to figure out how to do it in our case.

PR #593, opened 2018-08-24
Fix invalid chunk size in mbuf_alloc()

rprevot:
The mbuf_alloc() function stores an invalid chunk size in mbuf structure. 
This leads to SIGABRT during mbuf_dealloc() execution.


smukil (approved):
Thanks for the patch. Looks correct now.

ipapapa (approved):

PR #594, opened 2018-08-24
Avoid implicit interpretation of return type for dyn_aes_encrypt_msg()

smukil:
dyn_aes_encrypt_msg() returns a type of rstatus_t, however the function
returns a status only on an error and returns a 'count' on success
which the caller implicitly interprets as '# of bytes'. This pattern
is error prone and could lead to hard to track down bugs.

This patch also returns the right error code on the OOM case and also
adds checking on the caller side for this new error code DN_ENOMEM.

This was noticed while debugging a core dump. While this patch is only
for clean up, a future patch will change some more code in this area
to fix a crash bug.

### DIFF
--- a/src/dyn_crypto.c
+++ b/src/dyn_crypto.c
@@ -388,12 +388,13 @@ dyn_aes_decrypt(unsigned char *enc_msg, size_t enc_msg_len, struct mbuf *mbuf, u
  *
  */
 rstatus_t
-dyn_aes_encrypt_msg(struct msg *msg, unsigned char *arg_aes_key)
+dyn_aes_encrypt_msg(struct msg *msg, unsigned char *arg_aes_key, size_t* outlen)
 {
     struct mhdr mhdr_tem;
     int count = 0;
 
     if (STAILQ_EMPTY(&msg->mhdr)) {
+        // 'msg' is empty. Nothing to encrypt.
         return DN_ERROR;
     }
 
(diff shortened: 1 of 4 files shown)


shailesh33:
Thank you. Theres also some notion about error codes. There are some error codes for dynomite and some from redis and some from system. May be unification of these will be helpful?

shailesh33:
Sorry, I will still be poking my nose around this repo :)

shailesh33:
Looks good

smukil:
> Sorry, I will still be poking my nose around this repo :)

Please do! Would much appreciate your input/guidance whenever possible.

I think that we should take actions based on the Redis error codes only in Redis specific functions, which I think mostly live in src/proto/dyn_redis.c.

Otherwise those errors should probably be sent back directly to the client, or we should interpret it to a Dynomite equivalent error code if we want to take actions in other parts of the code that are not Redis specific.

Eg: If Redis returns OOM, then we should reinterpret that as DN_ENOMEM _if_ we want to take some action based on that in parts of the code that are not redis specific. I've not gone deep in the code to know whether this is already happening or not, but that's my general take on it.

So to summarize, I think your suggestion of unification from a general Dynomite point of view will be helpful. So we'll have generic code to deal with failures from memcache or Redis (or any other storage engine that we may support) in the same way, given that they're not engine specific errors.

ipapapa (approved):

PR #596, opened 2018-08-24
Fix mbuf usage in dnode_peer_handshake_announcing

rprevot:
When gossip is enabled, the "[dnode_peer_gossip_forward](https://github.com/Netflix/dynomite/blob/e63924fe4b186ab0d50927f2c832d03a1ea72d16/src/dyn_dnode_request.c#L191)" function is called periodically to spread out data. That function takes ownership of the "mbuf" parameter: "mbuf" is inserted into forwarded message or put back in mbuf pool.

More specifically, that function is called by "[dnode_peer_handshake_announcing](https://github.com/Netflix/dynomite/blob/e63924fe4b186ab0d50927f2c832d03a1ea72d16/src/dyn_dnode_peer.c#L552)". In that case, the same "mbuf" is passed on each gossip forward:
```
rstatus_t
dnode_peer_handshake_announcing(void *rmsg)
{
    //....
    struct mbuf *mbuf = mbuf_get();
    //....

    //for each peer, send a registered msg
    for (i = 0; i < nelem; i++) {
        //.....
        dnode_peer_gossip_forward(sp->ctx, conn, mbuf);
    }

    //free this as nobody else will do
    //mbuf_put(mbuf);

    return DN_OK;
}
```
"mbuf" instance is thus owned by multiple messages. This brings some inconsistencies. E.g. release of these messages ([msg_put](https://github.com/Netflix/dynomite/blob/e63924fe4b186ab0d50927f2c832d03a1ea72d16/src/dyn_message.c#L619) calls) will trigger "mbuf" recycling multiple times.

To avoid such scenario, this PR uses the "mbuf" instance declared in "dnode_peer_handshake_announcing" as a prototype. A new copy of the prototype is passed to "dnode_peer_gossip_forward" function at each iteration.

### DIFF
--- a/src/dyn_dnode_peer.c
+++ b/src/dyn_dnode_peer.c
@@ -529,12 +529,14 @@ dnode_peer_forward_state(void *rmsg)
     if (conn == NULL) {
         //running out of connection due to memory exhaust
         log_debug(LOG_ERR, "Unable to obtain a connection object");
+        mbuf_put(mbuf);
         return DN_ERROR;
     }
 
     status = conn_connect(sp->ctx, conn);
     if (status != DN_OK ) {
         conn_close(sp->ctx, conn);
+        mbuf_put(mbuf);
         log_debug(LOG_ERR, "Error happened in connecting on conn %d", conn->sd);
         return DN_ERROR;
     }
@@ -602,6 +604,7 @@ dnode_peer_handshake_announcing(void *rmsg)
         if (conn == NULL) {
             //running out of connection due to memory exhaust
             log_debug(LOG_DEBUG, "Unable to obtain a connection object");
+            mbuf_put(mbuf);
             return DN_ERROR;
         }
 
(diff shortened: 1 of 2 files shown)


smukil:
Thanks for the contribution. It seems like a correct fix, but I'm not sure about the implication on performance yet, so I'll let someone who's worked on the project for a bit longer to take a look.

For discussion's sake, I wonder if we should introduce a notion of reference-counted read only buffers. That will help us avoid these kinds of copies.

smukil:
After spending some more time thinking about this, we'd need to make quite a large change to do this the right way _and_ avoid the copies.

Just to document my thinking somewhere, my aim is to have a clear definition of who allocates and who frees buffers like mbuf and msg structs, and document the same in code. While digging into this, I found that this would require changing many parts of the code since hot functions like all the *_recv_next() and *_send_next() functions will need to change, which in turn would require all the callers of those functions to be adjusted accordingly.

So my take is that this can go in for now to unblock this use case, and we can do a larger change in the future to address the copies separately.

smukil (approved):

PR #597, opened 2018-08-28
Consistency for SMEMBERS when using DC_SAFE_QUORUM

smukil:
Although the data written to the same set in different replics will be
the same, the order in which they're returned in each replica might
be different. This causes the checksums of each replica set to be
different, causing SMEMBERS to fail under DC_SAFE_QUORUM.

This patch addresses this by rewriting all SMEMBERS calls to the
following (only when DC_SAFE_QUORUM is set):
 * SMEMBERS <set> -> SORT <set> ALPHA

This ensures that the returned set is in the same order across replicas
and hence ensures that their checksums match.

A mini-framework for rewriting queries is introduced to make future
query rewrites easier. Note that this is different from query
fragmenting which already exists today for queries like MGET.

### DIFF
--- a/src/proto/dyn_redis.c
+++ b/src/proto/dyn_redis.c
@@ -369,6 +369,99 @@ redis_error(struct msg *r)
 }
 
 /*
+ * Detects the query and does a rewrite if applicable.
+ *
+ * Currently the following queries are rewritten:
+ * 1) SMEMBERS <set> -> SORT <myset> ALPHA (only when DC_SAFE_QUORUM=true)
+ *    We rewrite this query this way since when DC_SAFE_QUORUM is enabled,
+ *    we run this query on multiple nodes and take checksums and compare that
+ *    they're the same. Since SMEMBERS offers no ordering guarantee, even though
+ *    the elements are the same, the order of elements in the set might be
+ *    different causing the checksums to be different and hence causing the
+ *    query to fail. Rewriting it to a SORT query ensures ordering and thus
+ *    ensures that the checksum comparison succeeds.
+ *
+ * * Sets *did_rewrite='true' if a rewrite occured and 'false' if not.
+ * * Does not modify 'orig_msg' and sets 'new_msg_ptr' to point to the new 'msg' struct with the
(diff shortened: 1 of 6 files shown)


smukil:
All testing done so far was on a local machine. I will test on a live cluster soon and post the results as well.

src/proto/dyn_redis.c
```diff
@@ -368,6 +368,99 @@ redis_error(struct msg *r)
...
+ *    different causing the checksums to be different and hence causing the
+ *    query to fail. Rewriting it to a SORT query ensures ordering and thus
+ *    ensures that the checksum comparison succeeds.
+ *
+ * * Sets *did_rewrite='true' if a rewrite occured and 'false' if not.
+ * * Does not modify 'orig_msg' and sets 'new_msg_ptr' to point to the new 'msg' struct with the
+ *   rewritten query if 'did_rewrite' is true.
+ * * Caller must take ownership of the newly allocated msg '*new_msg_ptr'.
+ */
+rstatus_t redis_rewrite_query(struct msg* orig_msg, struct context* ctx, bool* did_rewrite,
```
ipapapa, line 389: Shouldn't the function name be related to the `SMEMBERS` call? Assuming we want to rewrite others, later on, then we cannot reuse this function as it is directly related to `SMEMBERS`.

smukil: The function itself is meant to be used for any Redis query rewrite. We most likely wouldn't need more parameters for future rewrites. The body of the function has a switch case which currently only checks for MSG_REQ_REDIS_SMEMBERS. As more rewrites are required, we'll add more cases to this switch statement.

ipapapa (approved):

smukil:
Tested on a live cluster and it works as expected.

Without my patch, running SMEMBERS <set> with DC_SAFE_QUORUM says:
(error) ERR Dynomite: Failed to achieve Quorum

And with my patch we get the set back. I tried small and large sets (up to 4000 elements) to ensure testing multi-buffer responses as well.

Merging this.

PR #603, opened 2018-08-31
Auto-format the codebase to have consistent formatting

smukil:
I noticed that there were quite a few formatting inconsistencies in
the codebase where a single standard didn't seem to be followed.

I used a tool that I've used in other open source projects before
called clang-format-diff which automatically formats the entire
codebase with a single consistent formatting style.

The command I ran to format the entire codebase:
git diff $(printf '' | git hash-object -t tree --stdin) src/ | clang-format-diff-5.0 -style Google -p1 -i

For all future commits, we can run the following command to format only
the commit being submitted:

git diff -U0 HEAD^ | clang-format-diff-5.0 -style Google -p1 -i

Notably, this changes the tab space to 2 and reorders the includes
alphabetically. I found a few bugs due to the include reorders and
fixed them as well.


smukil:
This might be seen as a controversial change since it changes the fundamental formatting style, so I'll let folks weigh in.

This does not have any behavioral change and is purely mechanical formatting, except for the few include order bugs found and fixed.

ipapapa:
I am kind of favor of this PR. My logic is (a) formatting consistency is a good thing; (b) it helps the primary developer not to think about formatting inconsistencies. However, I would argue that these formatting changes needs to be limited to .c and .h files of the project. I have not seen if other are affected but this is something that @smukil can verify.

ipapapa (approved):

ipapapa:
I also think that #596 should be given priority over this PR. Because we will have re-update that branch continuously.

shailesh33:
Looks good. Does vi or other editors support this format easily? is there a settings file for vi/emacs that can be imported for this clang format diff tool?

smukil:
@ipapapa Thanks. Yes, it only applies to .c and .h files. Also, I will take a look at #596 first before merging this.

@shailesh33 Yes, I believe Google's own vimrc for C++ applies to C as well. Here it is:
https://github.com/vim-scripts/google.vim/blob/master/indent/google.vim

shailesh33:
sounds good.

smukil:
Rebased the patch on top of conflicting changes that went in recently. Merging now.

Issue #436, opened 2017-04-20
sysconfdir not honored by configure script


breauxaj:
Trying to build the project for CentOS as an RPM and when executing the configure script with:

./configure \
  --prefix=%{_sysconfdir}/%{name} \
  --exec-prefix=%{_prefix} \
  --bindir=%{_bindir} \
  --sbindir=%{_sbindir} \
  --mandir=%{_mandir} \
  --libdir=%{_libdir} \
  --sysconfdir=%{_sysconfdir}/%{name}

The values never seem to propagate to the resulting binary. The error reported:

conf_open:1664 conf: failed to open configuration 'conf/dynomite.yml': No such file or directory

If _sysconfdir was being used, it should be /etc/dynomite and not simply conf as the error reports.

shailesh33:
@breauxaj how are you running dynomite?

Since this project as been adapted and I lack the expertise in autoconf, and other build process, but I am investigating on how to do this with minimal regression or whether it should be done at all. I see I need to add 
AM_CPPFLAGS += -DSYSCONFDIR='$(sysconfdir)'
to the makefile.am and use that in the code. I also see the conf file is default to conf/dynomite.yml. If you provide the conf file as -c arg, it should work.

breauxaj:
The application builds successfully but assumes it will be deployed as is instead of packaged. This limits its adoption. I'm proposing that fixing these issues will increase it's potential audience.

If I run the application after build, it finds the config file, etc since its still in the build tree but that's not the way to deploy software. I'm following the RedHat model for deploying application software, putting the binary in /usr/bin, putting the configuration in /etc/dynomite, etc. By supporting the parameters, a builder can package this consistently.

Felixoid:
Hello. I've done my best to implement the functionality from the topic without a backward compatibility breaking in #604. Could you take a look, please, @ipapapa or @smukil ?

PR #604, opened 2018-09-12
Add support for sysconfdir for default config place

Felixoid:
By running of `./configure --sysconfdir="/etc/dynomite" --enable-packaging` default config will be another for now:
```
>±> ./src/dynomite -h  2>&1| grep conf-file
  -c, --conf-file=S            : set configuration file (default: /etc/dynomite/dynomite.yml)
```

This PR is related to #436

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -194,19 +194,6 @@ AS_IF([test "x$disable_stats" = xyes],
   [AC_DEFINE([HAVE_STATS], [1], [Define to 1 if stats is not disabled])])
 AC_MSG_RESULT($disable_stats)
 
-AC_ARG_ENABLE([packaging],
-  [AS_HELP_STRING(
-    [--enable-packaging],
-    [apply packaging standards])
-  ],
-  [enable_packaging=yes],
-  [enable_packaging=no])
-AS_IF(
-  [test "x$enable_packaging" = xyes],
-  [AC_DEFINE([PACKAGING], [1], [Define to 1 if packaging is enabled])],
-  [])
-AC_MSG_RESULT($enable_packaging)
-
 # Untar the yaml-0.1.4 in contrib/ before config.status is rerun
 AC_CONFIG_COMMANDS_PRE([tar xvfz contrib/yaml-0.1.4.tar.gz -C contrib])
 
--- a/configure.ac
+++ b/configure.ac
@@ -194,6 +194,19 @@ AS_IF([test "x$disable_stats" = xyes],
   [AC_DEFINE([HAVE_STATS], [1], [Define to 1 if stats is not disabled])])
 AC_MSG_RESULT($disable_stats)
 
+AC_ARG_ENABLE([packaging],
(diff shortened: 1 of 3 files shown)


smukil:
Thanks for the contribution @Felixoid. I'll have a look at your change soon. I'm new to the project, so I'm still unfamiliar with the build process, so I'll spend some time understanding it before I review this patch.

Felixoid:
Thank you, sure. Don't hesitate to ask, if something will be unclear.

smukil:
@Felixoid Why aren't the other dirs added too, that we'd ideally want to honor while enabling packaging? Eg: prefix, bindir, mandir, etc.

Felixoid:
Heh, pretty good question.

You are right, I'll add them as well

Felixoid:
So, after a little refresh, I see, that building process behave properly for other dirs while `./configure`. You could take a [PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=dynomite#n31) as an example.

There is the result of building:
```
>±> pacman -Ql dynomite 
dynomite /usr/
dynomite /usr/bin/
dynomite /usr/bin/dynomite
dynomite /usr/bin/dynomite-hash-tool
dynomite /usr/bin/dynomite-test
dynomite /usr/share/
dynomite /usr/share/licenses/
dynomite /usr/share/licenses/dynomite/
dynomite /usr/share/licenses/dynomite/LICENSE
dynomite /usr/share/man/
dynomite /usr/share/man/man8/
dynomite /usr/share/man/man8/dynomite.8.gz
```

smukil:
@Felixoid Thanks for verifying that. I think we should have one of 2 behaviors:

1. There isn't an '--enable-packaging' flag, but just specifying '--sysconfdir=<...>' or '--sbindir=<...>' is enough to have the values picked up.
2. None of the 'sysconfdir' or 'sbindir' values are picked up even if they're specified, _unless_ '--enable-packaging' is used.

Having the behavior where '--enable-packaging' is used only to pick up the 'sysconfdir' value can be quite confusing to users without context from this PR. What do you think?

Felixoid:
Here is some problem... Current behavior doesn't support change for `DN_CONF_PATH`. I would like to make it always `syconfdir` by default. Then it will be impossible to run just `dynomite` binary without `-c` argument, but, IMHO, it's still more convinient.

So, tl;dr: remove `--enable-packaging` and make `DN_CONF_PATH` always dependent on sysconfdir (or, if empty, make it as now equal to `conf/dynomite.yml`)

Felixoid:
Like this. I'll take a look on tests if they need to be adapted.

If you like this behavior, then I'll clean up my branch

Felixoid:
Hello, any update?

smukil:
This looks reasonable. Thanks for making the change @Felixoid! Merging this.

smukil (approved):

PR #606, opened 2018-09-22
Error responses should contain mbufs with appropriate error strings

smukil:
We've noticed that when the outstanding message queue for a connection
(conn->omsg_q) reaches the MAX_CONN_QUEUE_SIZE, we set the
ENOTRECOVERABLE state. This eventually leads a server to call
server_ack_err() which sends a error response message (struct msg*)
back to the client. This message however does not contain a mbuf
with an error string, causing failures in the client.

Eg: The dyn_dnode_client tries to encrypt this empty message in
dnode_rsp_send_next(), which returns an error and eventually
leads to a crash later on.

This patch fixes this by filling in these error responses with
a valid mbuf which contains an error string.

No further cleanup is done here, however, some clean up needs to
be done around this code in the future.

### DIFF
--- a/src/dyn_dnode_peer.c
+++ b/src/dyn_dnode_peer.c
@@ -302,7 +302,13 @@ static void dnode_peer_ack_err(struct context *ctx, struct conn *conn,
   // Create an appropriate response for the request so its propagated up;
   // This response gets dropped in rsp_make_error anyways. But since this is
   // an error path its ok with the overhead.
-  struct msg *rsp = msg_get(conn, false, __FUNCTION__);
+  struct msg *rsp = msg_get_error(conn, PEER_CONNECTION_REFUSE, conn->err);
+  if (rsp == NULL) {
+    // TODO: It's not clear how the peer should behave if we hit this error
+    // condition. Return an appropriate error instead.
+    log_warn("Could not allocate msg for notifying an error to peer");
+    return;
+  }
   req->done = 1;
   rsp->peer = req;
   rsp->is_error = req->is_error = 1;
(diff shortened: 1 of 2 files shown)


src/dyn_dnode_peer.c
```diff
@@ -302,7 +302,12 @@ static void dnode_peer_ack_err(struct context *ctx, struct conn *conn,
   // Create an appropriate response for the request so its propagated up;
   // This response gets dropped in rsp_make_error anyways. But since this is
   // an error path its ok with the overhead.
-  struct msg *rsp = msg_get(conn, false, __FUNCTION__);
+  struct msg *rsp = msg_get_error(conn, PEER_CONNECTION_REFUSE, conn->err);
+  if (rsp == NULL) {
+    // TODO: This is wrong. Return an appropriate error instead.
```
ipapapa: I am having a hard time understanding the comment `this is wrong`.

smukil: It should be clear how every function should behave under successful cases and under error conditions. In this function, the server is attempting to tell the client that an error has occurred. When this function fails to allocate a message, it just adds a logline and does nothing else. So, the client would never know what happened to its request, and how the client should behave in this case is undefined.

Ideally functions like this should never fail, but if there's no way to make them not fail, the client should have some logic to anticipate this case as well.

ipapapa: I get it, but the code comment was not clear when I read it (and it is probably not clear outside of the context of this PR). My recommendation is to add more context but not mandatory.

ipapapa (approved):

src/dyn_dnode_peer.c
smukil: Good point. I updated the comment to make it more clear. Thanks for the review!

PR #610, opened 2018-09-30
can compile openssl 1.1.0 

charsyam:
current encrypt/decrypt code is based on openssl 1.0.2.
but from openssl 1.1.0, the interface is changed.

this pr can compile openssl 1.1.0 and higher version.

it passed dynomite-test


charsyam:
oh I couldn't see latest version. sorry.

PR #612, opened 2018-10-01
Merge dev into v0.6

smukil:
Includes fix for crash in full debug mode.

### DIFF
--- a/src/dyn_core.h
+++ b/src/dyn_core.h
@@ -23,64 +23,6 @@
 #ifndef _DYN_CORE_H_
 #define _DYN_CORE_H_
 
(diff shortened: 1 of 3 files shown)


shailesh33:
Just curios, did this always exist or was a side effect of removing cyclic dependencies?

smukil:
@shailesh33 It was a side effect of removing the cyclic dependencies. After that change, one of the files failed to include the header that had a lot of the compile time values initialized, which included the flag which determined whether we were in debug mode or not.

The verbose explanation is in the commit message here:
https://github.com/Netflix/dynomite/pull/612/commits/fe5667bd39773f13ec7f8d2ab0429fc0d33a1026

PR #614, opened 2018-10-12
Unnest core initialization functions and add extra error checking

smukil:
The core_start() function intializes everything required for dynomite
to be setup correctly. However, the code had nested all the functions
within core_ctx_create(), which recursively initialized unrelated
structures.

This patch unnests the entire chain which allows for better readability
and it also adds some extra error checking to avoid segfaults when
we fail to allocate memory, etc.

### DIFF
--- a/src/dyn_core.c
+++ b/src/dyn_core.c
@@ -84,24 +84,18 @@ static rstatus_t core_gossip_pool_init(struct context *ctx) {
   CBUF_Init(C2G_OutQ);
 
   THROW_STATUS(gossip_pool_init(ctx));
-  THROW_STATUS(core_init_last(ctx));
   return DN_OK;
 }
 
 static rstatus_t core_dnode_peer_pool_preconnect(struct context *ctx) {
   rstatus_t status = dnode_peer_pool_preconnect(ctx);
   IGNORE_RET_VAL(status);
-  status = core_gossip_pool_init(ctx);
-  // if (status != DN_OK)
-  //  gossip_pool_deinit(ctx);
   return status;
 }
 static rstatus_t core_dnode_peer_init(struct context *ctx) {
   /* initialize peers */
   THROW_STATUS(dnode_initialize_peers(ctx));
-  rstatus_t status = core_dnode_peer_pool_preconnect(ctx);
-  if (status != DN_OK) dnode_peer_pool_disconnect(ctx);
-  return status;
+  return DN_OK;
 }
 
 static rstatus_t core_dnode_proxy_init(struct context *ctx) {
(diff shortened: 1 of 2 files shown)


shailesh33:
ok ok. I agree nesting is not the best of the thing but the general idea is to have a chain of initializers and deinitializers which at any given failure will unwind and uninitialize as it goes. Is there a specific order in which deinit should happen?

smukil:
> ok ok. I agree nesting is not the best of the thing but the general idea is to have a chain of initializers and deinitializers which at any given failure will unwind and uninitialize as it goes. Is there a specific order in which deinit should happen?

Thanks for taking a look! Yes, I agree with that. I've made sure to de-initialize in the reverse order as the initialization is done. The "error:" label also has a comment stating the same under it.

Also, I've checked that we won't be accessing any invalid or uninitialized struct/variable if an error is hit early on in the initialization chain.

shailesh33:
> > ok ok. I agree nesting is not the best of the thing but the general idea is to have a chain of initializers and deinitializers which at any given failure will unwind and uninitialize as it goes. Is there a specific order in which deinit should happen?
> 
> Thanks for taking a look! Yes, I agree with that. I've made sure to de-initialize in the reverse order as the initialization is done. The "error:" label also has a comment stating the same under it.
> 
> Also, I've checked that we won't be accessing any invalid or uninitialized struct/variable if an error is hit early on in the initialization chain.

It will be good if we can do this is such a way that it doesnt depend on developer to do the right thing. May be a table of init and deinit functions. Just a thought.
The PR per se looks good to me. Feel free to merge if you do not want to take it that far.

smukil:
Thanks for the suggestion @shailesh33. That seems like a good idea. To go a bit further, if we're able to establish a hierarchy of structures each with its own setup() and destroy() functions (similar to constructors/destructors in C++), when anything goes wrong, we can call the top level struct's destroy() function, which would recursively destroy all the initialized nested structs. (The top-level struct will basically be the table of structs to init as you mentioned.)

That way, the onus isn't on the developer to do the right thing, and the code will be simpler to understand.

This is a much larger change though, but I think we can do it in stages.

shailesh33:
> Thanks for the suggestion @shailesh33. That seems like a good idea. To go a bit further, if we're able to establish a hierarchy of structures each with its own setup() and destroy() functions (similar to constructors/destructors in C++), when anything goes wrong, we can call the top level struct's destroy() function, which would recursively destroy all the initialized nested structs. (The top-level struct will basically be the table of structs to init as you mentioned.)

They need not be nested structures. They can be components in the system which depend/use each other for example deinit msg and mbuf should be last. 
I agree its a larger change. Go ahead and merge this PR.

shailesh33 (approved):

smukil:
@shailesh33 You're right, that's what I meant. I didn't mean to imply that all the structs need to be nested under a top level struct, but more like structs can refer to other structs, etc. Thanks for the review!

PR #617, opened 2018-11-12
Refactor test framework to easily expand on cluster management

smukil:
This patch takes a lot of the code from test/cluster_generator.py
and refactors it into easily manageable classes which allows for
easily creating and destroying a dynomite cluster in a local test
environment. Refer to the last paragraph of this commit message for
the motivation behind this patch.

To create a dynomite cluster, now it's as simple as doing the following:

  cluster = DynoCluster(args.request_file, ips)
  cluster.launch()
  \# Run all the required tests
  cluster.teardown()

This patch does this by forming clear relations between DynoCluster,
DynoSpec, DynoNode and RedisNode.

DynoCluster --manages--> list(DynoSpec), list(DynoNode)
DynoNode    --manages--> RedisNode
DynoNode and DynoSpec have a 1:1 relationship.

In the future, we should decouple RedisNode and DynoNode and abstract
away specific storage node classes to a generic class.
Eg: 'RedisNode' implements 'StorageNode', and 'DynoNode' directly
uses only 'StorageNode'.

This patch also fixes a small bug where the incorrect port is used
for the client listening port.

This patch currently breaks other standalone scripts like func_test.py
and large_value_test.py. This will be cleaned up in a future commit,
and IMO it's not a serious issue as it's difficult to already use
those scripts since they assume that a cluster with a very specific
configuration is already setup (which is not that simple to do
manually today).

The initial motivation behind this patch is to create scripts to
_only start_ or _only stop_ a dynomite cluster for the purpose
of manual testing against a cluster. Before this patch,
cluster_generator.py always cleaned up the resources on exit. Future
commits will introduce new scripts like start_cluster.py and
teardown_cluster.py which will allow developers to just start a cluster
and run tests against it and then tear it down later. Other things
that this patch enables is to do hypothetical experiments like
dual running against 2 dynomite clusters with different
configurations, expanding the framework to run only a subset of tests
based on user flags passed to cluster_generator.py, etc.

Future TODO: Add code comments to classes and methods as necessary.

### DIFF
--- a/test/cluster_generator.py
+++ b/test/cluster_generator.py
@@ -1,34 +1,23 @@
 #!/usr/bin/env python3
 from contextlib import ExitStack
-from tempfile import mkdtemp
 from time import sleep
 from urllib.request import urlopen
 import argparse
 import json
 import random
 import sys
+import yaml
 
-from plumbum import LocalPath
 from plumbum import local
 
 from dyno_cluster import DynoCluster
 from func_test import comparison_test
-from utils import generate_ips
+from utils import generate_ips, setup_temp_dir
 from redis_node import RedisNode
 
 REDIS_PORT = 1212
 STATS_PORT = 22222
 SETTLE_TIME = 5
-
-def setup_temp_dir():
-    tempdir = LocalPath(mkdtemp(dir='.', prefix='test_run.'))
-    (tempdir / 'logs').mkdir()
-    confdir = (tempdir / 'conf')
-    confdir.mkdir()
-
-    LocalPath('../../conf/dynomite.pem').symlink(confdir / 'dynomite.pem')
-
-    return tempdir
 
 def main():
     parser = argparse.ArgumentParser(
--- a/test/cluster_generator.py
+++ b/test/cluster_generator.py
@@ -1,188 +1,24 @@
 #!/usr/bin/env python3
-from collections import namedtuple
 from contextlib import ExitStack
-from contextlib import contextmanager
-from signal import SIGINT
 from tempfile import mkdtemp
 from time import sleep
 from urllib.request import urlopen
 import argparse
 import json
 import random
-import socket
 import sys
-import yaml
 
-from plumbum import FG
-from plumbum import BG
 from plumbum import LocalPath
 from plumbum import local
 
-from ip_util import quad2int
-from ip_util import int2quad
 from dyno_cluster import DynoCluster
-from dyno_node import DynoNode
 from func_test import comparison_test
+from utils import generate_ips
 from redis_node import RedisNode
 
-DYN_O_MITE_DEFAULTS = dict(
-    secure_server_option='datacenter',
-    pem_key_file='conf/dynomite.pem',
-    data_store=0,
-    datastore_connections=1,
-)
-INTERNODE_LISTEN = 8101
-CLIENT_LISTEN = 8102
 REDIS_PORT = 1212
 STATS_PORT = 22222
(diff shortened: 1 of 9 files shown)


ipapapa (approved):

test/utils.py
```diff
@@ -1,9 +1,69 @@
...
+        (dc['name'], tokens_for_dc(dc['racks']))
+        for dc in dcs
+    ]
+
+
+def dc_count(dc):
+    return sum(count for rack, count in dc)
+
+def generate_ips():
+    state = BASE_IPADDRESS
```
hashbrowncipher: Hmm this could be itertools.count.

(Reviewing my own code, I think)

test/utils.py
```diff
@@ -1,9 +1,69 @@
...
     return ''.join(random.choice(chars) for _ in range(size))
 
 def number_generator(size=4, chars=string.digits):
     return ''.join(random.choice(chars) for _ in range(size))
+
+def dict_request(request):
+    """Converts the request into an easy to consume dict format.
+
+    We don't ingest the request in this format originally because dict
+    ordering is nondeterministic, and one of our goals is to deterministically
```
hashbrowncipher: I think this is no longer true in Python 3.6. I wonder if we should adapt the code to newer versions of Python.

test/dyno_node.py
```diff
@@ -2,16 +2,42 @@
...
+from plumbum import BG
+from plumbum import local
+
+dynomite_bin = local.get('./test/_binaries/dynomite', 'src/dynomite')
 
 class DynoNode(Node):
-    def __init__(self, ip, port=8102,
-                 dnode_port=8101, data_store_port=1212):
-        super(DynoNode, self).__init__(ip, port)
+    '''TODO: Decouple from Redis and make it more generic'''
```
hashbrowncipher: This kind of comment is typically prefaced with `#`, not `"""`. `''''` is valid, but rare in Python syntax.

If you want, you could download code formatter like [black](https://pypi.org/project/black/). It will do all this for you.

test/cluster_generator.py
```diff
@@ -202,26 +37,40 @@ def main():
...
-            redis_info = stack.enter_context(launch_redis(standalone_redis_ip))
-            dynomite_info = stack.enter_context(launch_dynomite(nodes))
+    # Create a Dynomite cluster.
+    dynomite_cluster = DynoCluster(args.request_file, ips)
 
+    # Make sure to change the working directory to the temp dir before running the tests.
+    with local.cwd(temp):
+        try:
+            # Launch the standalone Redis node and the dynomite cluster.
+            standalone_redis.launch()
```
hashbrowncipher: Unless there's a reason not to, I think it's worth preserving this as a context manager API. Let's discuss it in person.

test/utils.py
smukil: Done.

test/utils.py
smukil: I think I'll hold off on this for now since I don't want to introduce a minimum Python version requirement at this point.

test/dyno_node.py
smukil: Done.

test/cluster_generator.py
smukil: Thanks for this pointer. Makes the code much cleaner.

smukil:
I'm merging this to unblock myself. If anyone has more review comments for this in the future, I'll address them in a separate patch.

PR #626, opened 2019-01-10
Support remaining SCRIPT commands. (EXISTS, KILL, FLUSH)

smukil:
This patch adds support for the remaining SCRIPT commands.

SCRIPT FLUSH will clear the script caches of all the Redis instances
managed by Dynomite.

SCRIPT KILL will kill ANY script running on ANY instance of Redis
managed by Dynomite. This could prove tricky for multi-tenant use
cases where multiple users may be executing multiple scripts from
different nodes. Under this rare case, ALL running scripts will be
terminated.

SCRIPT EXISTS works as one would expect from Redis.

Tests have been added to test all the above except SCRIPT KILL. It's
tricky to write an automated test for this, but it has been manually
tested.

A new method is introduced to the test framework to allow execution of
commands only through Dynomite as opposed to always running it against
Redis and Dynomite.

### DIFF
--- a/src/dyn_message.h
+++ b/src/dyn_message.h
@@ -236,6 +236,9 @@ typedef enum msg_parse_result {
   ACTION(SENTINEL)                                                             \
   ACTION(REQ_REDIS_SCRIPT)                                                     \
   ACTION(REQ_REDIS_SCRIPT_LOAD)                                                \
+  ACTION(REQ_REDIS_SCRIPT_EXISTS)                                              \
+  ACTION(REQ_REDIS_SCRIPT_FLUSH)                                               \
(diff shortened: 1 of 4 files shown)


ipapapa (approved):

ipapapa:
I feel we also need to update the Redis documentation: https://github.com/Netflix/dynomite/blob/dev/notes/redis.md

smukil:
> I feel we also need to update the Redis documentation: https://github.com/Netflix/dynomite/blob/dev/notes/redis.md

Oops, I made the edit but forgot to add the changes to this commit. I posted a new commit for this:
https://github.com/Netflix/dynomite/pull/628

I've also merged the Redis Lua scripting wiki page into the Redis commands wiki page.

PR #629, opened 2019-01-15
Remove code duplication in dyn_client, more code cleanup and misc bug…

smukil:
… fixes

There were multiple code paths with duplicated code to do the same
thing, namely copy the 'msg' and forward it to the right peer. This
patch introduces a new funciton req_forward_to_peer() which
encapsulates this functionality once.

This also fixes a bug which previously only compared rack strings to
figure out if they are the same rack, which could be wrong if racks
across DCs have the same name.

There are also multiple other small code cleanup changes including
more code comments.

A small but critical bug is fixed in msg_clone().

### DIFF
--- a/src/dyn_client.c
+++ b/src/dyn_client.c
@@ -512,7 +512,7 @@ static void req_forward_stats(struct context *ctx, struct msg *req) {
   }
 }
 
-rstatus_t local_req_forward(struct context *ctx, struct conn *c_conn,
+rstatus_t req_forward_local_datastore(struct context *ctx, struct conn *c_conn,
                             struct msg *req, uint8_t *key, uint32_t keylen,
                             dyn_error_t *dyn_error_code) {
   rstatus_t status;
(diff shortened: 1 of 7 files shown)


src/dyn_client.c
```diff
@@ -600,34 +600,93 @@ static rstatus_t admin_local_req_forward(struct context *ctx,
...
+    // We forward the error if the target was in the same rack, or if it was in a remote
+    // rack and we're expecting a quorum response.
+    if ((!same_rack && req->consistency != DC_ONE) || same_rack) {
+      req_forward_error(ctx, c_conn,
+          (rack_msg ? rack_msg : req),
+          status, *dyn_error_code);
+    }
+  }
+  // Release the copy if we made one above..
+  if (rack_msg != NULL && (force_copy || !same_rack)) req_put(rack_msg);
```
shailesh33: its up to you, but generally do not put the statement on the same line as the condition check

shailesh33 (approved):

src/dyn_client.c
smukil: Good point. Done.

ipapapa (approved):

PR #632, opened 2019-02-04
Record argument positions in query request

smukil:
Currently the Redis parser in Dynomite only records the position of
the initial key as a 'struct keypos', and passes along the remaining
arguments as is to Redis. This patch records the argument positions
as well during request parsing. This will allow us to easily refer back
to the arguments after the parsing stage for things like query rewrites,
further request analysis, etc. The newly introduced 'args' array is not
used in this patch, but will be used in upcoming patches motivated by
the read-repairs work.

This patch also renames the fields prefixed with "narg*" in 'struct msg'
to "ntoken*". This is because the code actually treats all 'narg's as
tokens and solely not as arguments. The rename is done to avoid
ambiguity between arguments and tokens.

### DIFF
--- a/src/dyn_message.c
+++ b/src/dyn_message.c
@@ -356,6 +356,12 @@ done:
     return NULL;
   }
 
+  msg->args = array_create(1, sizeof(struct argpos));
+  if (msg->args == NULL) {
+    dn_free(msg);
+    return NULL;
+  }
+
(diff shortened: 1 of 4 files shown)


ipapapa (approved):

ipapapa:
LGTM

PR #637, opened 2019-02-26
[WIP] Read repairs for primary keys

smukil:
This patch adds read repairs for primary keys. It specifically works
only with the following commands at the moment:
 - SET
 - GET
 - DEL
 - APPEND

Repiars need to be enabled under a flag in the YAML configuration. It is
enabled only if 'repairs_enabled: true' is specified.

This current implementation of repairs is not supported and returns the
values in a custom format:
1) E/N (exists or not-exists)
2) Value itself
3) Timestamp of value

This custom format will be removed in future patches to standardize on
the regular Redis wire protocol.

Currently the timestamp is generated on the server side. Client side
timestamp support will be added in future patches.

TODO:-
 - Add client side timestamp support.
 - Add repair support for secondary keys.
 - Remove custom result format.
 - Add support for tracking expired keys.
 - Add background repir process.
 - Add documentation on repair implementation.

### DIFF
--- a/src/proto/dyn_redis.c
+++ b/src/proto/dyn_redis.c
@@ -25,6 +25,7 @@
 
 #include "../dyn_core.h"
 #include "../dyn_dnode_peer.h"
+#include "../dyn_util.h"
 #include "dyn_proto.h"
 
 #define RSP_STRING(ACTION) ACTION(ok, "+OK\r\n")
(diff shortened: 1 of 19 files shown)


src/proto/dyn_redis.c
```diff
@@ -36,6 +37,35 @@
...
+
+struct cmd_info {
+  char cmd_str[20];
+  uint32_t num_args;
+  bool is_delete;
+};
+
+struct cmd_info redis_cmd_info[MSG_END_IDX];
+
+void redis_init_datastore() {
```
ipapapa, line 51: Do we have to do this for every single command?

src/proto/dyn_redis.c
```diff
@@ -375,6 +405,170 @@ rstatus_t record_arg(uint8_t* start_pos, uint8_t* end_pos, struct array *target_
...
 
+/*
+ * This is a generic script for writes to primary keys (SET, APPEND, DEL, etc.)
+ * The string is formatted to take arguments in the order specified below.
+ *
+ * FMT String order:-
+ * %d       %s    %s     %s     %d       %u   %d       %s   %d         %d      %d        %s
+ * <keylen> <key> <+set> <-set> <ts_len> <ts> <op_len> <op> <narg_len> <nargs> <val_len> <val>
+ */
+const char* REPAIR_WRITE_SCRIPT_FMT_STRING = "*10\r\n$4\r\nEVAL\r\n$720\r\n"
```
ipapapa, line 416: Instead of hard-coding the script can we load it from the YAML? If we also go through this path we need to have some error checking in case there are issues with the script.

src/proto/dyn_redis.c
```diff
@@ -375,6 +405,170 @@ rstatus_t record_arg(uint8_t* start_pos, uint8_t* end_pos, struct array *target_
...
+
+  // Make a copy of the key.
+  uint8_t *key_copy;
+  key_copy = (uint8_t*)strndup((char*)key, keylen);
+  key_copy[keylen] = '\0';
+
+  struct argpos *exists_pos = (struct argpos*)array_get(orig_msg->args, 0);
+  msg_type_t msg_type = (*(char*)exists_pos->start == 'E') ? MSG_REQ_REDIS_SET : MSG_REQ_REDIS_DEL;
+  switch(msg_type) {
+    case MSG_REQ_REDIS_SET:
```
ipapapa, line 554: Should we also have a `case` for the APPEND and the other command?

ipapapa:
We probably also need to update the README.md with the new property on the YAML. I am not sure if there are also entries on the wiki.

ipapapa:
We need to add in the PR description a link on how the read repairs are implemented e.g. a design document or some similar documentation.

smukil:
Abandoning this. New PR here:
https://github.com/Netflix/dynomite/pull/653

PR #653, opened 2019-04-15
[Unsupported] Read repairs implementation

smukil:
In order to repair mismatched replicas, we need to repair them based on some agreed
 upon value. In most systems that usually is based on the timestamps. Since Redis does
 not support timestamps, we need to add the logic that tracks timestamps as metadata.

 The way we achieve recording timestamps is by converting all writes to Lua scripts that
 atomically do the operation itself and store the metadata for that key.

 The metadata is stored in what we call 'add sets' and 'remove sets' which are basically
 sorted sets that store the key name along with a corresponding timestamp that denotes
 when that key was last updated. All primary keys go to a 'top level add/rem set' which
 is currently denoted by '._add-set' and '._rem-set'.

 Secondary keys (fields in hash maps, sets, etc.) are stored in add/rem sets of their
 corresponding primary keys and are denoted by '._add-set_<key>' and '._rem-set_<key>'.

 A majority of the logic lives in src/proto/dyn_redis_repair.c.

 The 'redis_cmd_info' static array of structures contains information that will be used
 by  the "rewrite with metadata" and "repair" logic. This array is indexed by the command
 index that is set in dyn_message.h.

 All queries that will be rewritten to include metadata are denoted by
 'redis_cmd_info[CMD_IDX].has_repair_support' where 'CMD_IDX' is the index of the
 corresponding query. All read queries that return the requested values along with the
 metadata are denoted by 'redis_cmd_info[CMD_IDX].is_repairable'.
 If queries do not have "rewrite with MD" or "repair" support, we fallback to the old
 behavior which is based on quorum checksums.

 Every repairable read ('is_repairable' == true) returns data in the Redis wire protocol
 in the following format:
 \<status> \<ts> \<value>

 , where status can be one of 'X', 'R' or 'E'. They stand for:
    X = value not found
    R = value was deleted (i.e. it was found in the rem-set)
    E = value exists (it was found in the add-set)

 This format was chosen because it's the most optimized in terms of returning values to
 the client, i.e. we can just move the start pointer of the buffer just before '<value>'
 by skipping '<status>' and '<ts>' thereby avoiding the need to copy the value to yet
 another buffer.

 Currently the only read commands that have repair support return only one value at a
 time. In the future when queries that can return multiple values will be returned,
 the format will be naturally extended to be:
 <status_1> <ts_1> <status_2> <ts_2> .. <status_n> <ts_n> <val_1> <val_2> .. <val_n>

 This will allow us to skip all the metadata with a simple pointer move just as we do
 now.

 On repair supported reads, if we find that the timestamps between responses from the
 responding replicas are not the same, we pick the replica that has the largest TS, take
 its value and send a new write command to the replicas that are out of date with the
 most updated value.

 Since we need to know the exact keys, fields and values while writing with metadata or
 while repairing, we do a "post-parsing" step that records these in order to efficiently
 populate the arguments for the Lua script that we will send to Redis on behalf of the
 read/write. (see post_parse_msg())
 This may change if we modify the parser itself to record these tokens efficiently.
 TODO: Support parsing of optional fields.

 The Lua scripts that we use as templates for supported commands are listed below as
 compile time constants.

 This is still in the beta stage and has some limitations:
 - Large values may cause overflows ( > 512 bytes)
 - Limited command support due to parser limitations:
     - Sorted sets and lists need optional command parsing support.
 - Repairing on the read path vs. the background has perf implications. In the future,
   all the heavy lifting will happen in the background in order to have a 0 perf
   overhead on the read and write paths.

### DIFF
--- a/src/dyn_conf.h
+++ b/src/dyn_conf.h
@@ -131,6 +131,9 @@ struct conf_pool {
   uint8_t datastore_connections;
   uint8_t local_peer_connections;
   uint8_t remote_peer_connections;
+
+  /* repairs enabled */
+  bool read_repairs_enabled;
 };
 
 struct conf {
--- a/src/proto/dyn_redis.c
+++ b/src/proto/dyn_redis.c
@@ -411,8 +411,6 @@ rstatus_t redis_rewrite_query(struct msg *orig_msg, struct context *ctx,
     case MSG_REQ_REDIS_SMEMBERS:
 
       if (orig_msg->owner->read_consistency == DC_SAFE_QUORUM) {
-        // SMEMBERS should have only one key.
-        ASSERT(orig_msg->nkeys == 1);
 
         // Get a new 'msg' structure.
         new_msg = msg_get(orig_msg->owner, true, __FUNCTION__);
--- a/src/proto/dyn_redis.c
+++ b/src/proto/dyn_redis.c
@@ -361,6 +361,20 @@ static bool redis_error(struct msg *r) {
   return false;
 }
 
+rstatus_t record_arg(uint8_t* start_pos, uint8_t* end_pos, struct array *target_arr) {
+  ASSERT(target_arr != NULL && start_pos != NULL && end_pos != NULL);
+
+  struct argpos *arg_pos;
+  arg_pos = array_push(target_arr);
+  if (arg_pos == NULL) {
+    return DN_ENOMEM;
+  }
+  arg_pos->start = start_pos;
+  arg_pos->end = end_pos;
+
+  return DN_OK;
+}
+
 /*
  * Detects the query and does a rewrite if applicable.
  *
@@ -549,17 +563,17 @@ void redis_parse_req(struct msg *r, const struct string *hash_tag) {
           }
           r->token = p;
           /* req_start <- p */
-          r->narg_start = p;
-          r->rnarg = 0;
+          r->ntoken_start = p;
+          r->rntokens = 0;
           state = SW_NARG;
         } else if (isdigit(ch)) {
-          r->rnarg = r->rnarg * 10 + (uint32_t)(ch - '0');
+          r->rntokens = r->rntokens * 10 + (uint32_t)(ch - '0');
         } else if (ch == CR) {
-          if (r->rnarg == 0) {
+          if (r->rntokens == 0) {
             goto error;
           }
-          r->narg = r->rnarg;
-          r->narg_end = p;
+          r->ntokens = r->rntokens;
+          r->ntoken_end = p;
           r->token = NULL;
           state = SW_NARG_LF;
         } else {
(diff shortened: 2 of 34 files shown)


ipapapa:
> On repair supported reads, if we find that the timestamps between responses from the
responding replicas are not the same, we pick the replica that has the largest TS, take
its value and send a new write command to the replicas that are out of date with the
most updated value.

If there has been a mutation after we have chosen the replica with the largest TS on read, do we compare the TS (again) on updating the value?

> Limited command support due to parser limitations:

Since we are going to have a follow-up PR for that, shall we create an issue to centrally track the progress of the feature?

src/dyn_conf.h
```diff
@@ -131,6 +131,9 @@ struct conf_pool {
   uint8_t datastore_connections;
   uint8_t local_peer_connections;
   uint8_t remote_peer_connections;
+
+  /* repairs enabled */
+  bool repairs_enabled;
```
ipapapa: shall we modify this to `read_repairs_enabled` to avoid confusion with a generic repair service feature?

src/proto/dyn_redis.c
```diff
@@ -407,12 +408,13 @@ rstatus_t redis_rewrite_query(struct msg *orig_msg, struct context *ctx,
...
   uint8_t *key = NULL;
   rstatus_t ret_status = DN_OK;
+
   switch (orig_msg->type) {
     case MSG_REQ_REDIS_SMEMBERS:
 
       if (orig_msg->owner->read_consistency == DC_SAFE_QUORUM) {
         // SMEMBERS should have only one key.
-        ASSERT(orig_msg->nkeys == 1);
+        //ASSERT(orig_msg->nkeys == 1);
```
ipapapa: Why was this `ASSERT` commented out?

src/proto/dyn_redis_repair.c
```diff
@@ -0,0 +1,1191 @@
...
+  int variadic_value_jump;
+  const char* rewrite_script;
+  msg_type_t repair_by_add;
+  msg_type_t repair_by_rem;
+};
+
+struct cmd_info redis_cmd_info[] = {
+  {"UNKNOWN", 0, 0, 0, 0, 0, false, false, false, -1, -1, -1, -1, -1, -1, NULL, 0, 0},
+
+  // Begin memcache commands. All Memcache commands are not supported. (idx 1 onwards)
```
ipapapa: We could potentially move the Memcache commands in a separate file related to Memcache.

src/dyn_conf.h
smukil: Good idea. Done.

src/proto/dyn_redis.c
smukil: I thought this was a minor issue in testing but I found out it has a larger impact later. I already merged this part here:
https://github.com/Netflix/dynomite/commit/1a5039997013949e9a89406fce9c6ca99da5bb87

I'll rebase over that.

src/proto/dyn_redis_repair.c
smukil: This needs to follow the same index order of the commands in dyn_message.c. And Redis and Memcache commands are listed as part of the same array there which is why we follow the same here.

I'll work on a larger patch to separate them out. For now, I've renamed the array to 'proto_cmd_info' and moved it to a new header file called "dyn_proto_repair.h"

smukil:
> If there has been a mutation after we have chosen the replica with the largest TS on read, do we compare the TS (again) on updating the value?


Yes, we do. The Lua script updates only if it doesn't see a TS later than the one it's trying to repair with.

> > Limited command support due to parser limitations:
> 
> Since we are going to have a follow-up PR for that, shall we create an issue to centrally track the progress of the feature?

Good point. I added it here:
https://github.com/Netflix/dynomite/issues/660

smukil:
Bad rebase, will push a different PR and close this.

PR #664, opened 2019-05-16
Fix conf parsing overflow

smukil:
After adding a new option to the 'struct conf_pool' list, a previously
undiscovered bug came to light where we were casting data types of
'short' to 'int' and 'bool' to 'int' which would give us the wrong
result. This patch fixes these bugs.


shailesh33 (approved):
looks good to me.

smukil:
@shailesh33 Thanks for the review!

PR #666, opened 2019-05-21
Read repairs beta with multibuf

smukil:
In order to repair mismatched replicas, we need to repair them based on some agreed
 upon value. In most systems that usually is based on the timestamps. Since Redis does
 not support timestamps, we need to add the logic that tracks timestamps as metadata.

 The way we achieve recording timestamps is by converting all writes to Lua scripts that
 atomically do the operation itself and store the metadata for that key.

 The metadata is stored in what we call 'add sets' and 'remove sets' which are basically
 sorted sets that store the key name along with a corresponding timestamp that denotes
 when that key was last updated. All primary keys go to a 'top level add/rem set' which
 is currently denoted by '._add-set' and '._rem-set'.

 Secondary keys (fields in hash maps, sets, etc.) are stored in add/rem sets of their
 corresponding primary keys and are denoted by '._add-set_<key>' and '._rem-set_<key>'.

 The 'redis_cmd_info' static array of structures contains information that will be used
 by  the "rewrite with metadata" and "repair" logic. This array is indexed by the command
 index that is set in dyn_message.h.

 All queries that will be rewritten to include metadata are denoted by
 'redis_cmd_info[CMD_IDX].has_repair_support' where 'CMD_IDX' is the index of the
 corresponding query. All read queries that return the requested values along with the
 metadata are denoted by 'redis_cmd_info[CMD_IDX].is_repairable'.
 If queries do not have "rewrite with MD" or "repair" support, we fallback to the old
 behavior which is based on quorum checksums.

 Every repairable read ('is_repairable' == true) returns data in the Redis wire protocol
 in the following format:
 <status> <ts> <value>

 , where status can be one of 'X', 'R' or 'E'. They stand for:
    X = value not found
    R = value was deleted (i.e. it was found in the rem-set)
    E = value exists (it was found in the add-set)

 This format was chosen because it's the most optimized in terms of returning values to
 the client, i.e. we can just move the start pointer of the buffer just before '<value>'
 by skipping '<status>' and '<ts>' thereby avoiding the need to copy the value to yet
 another buffer.

 Currently the only read commands that have repair support return only one value at a
 time. In the future when queries that can return multiple values will be returned,
 the format will be naturally extended to be:
 <status_1> <ts_1> <status_2> <ts_2> .. <status_n> <ts_n> <val_1> <val_2> .. <val_n>

 This will allow us to skip all the metadata with a simple pointer move just as we do
 now.

 On repair supported reads, if we find that the timestamps between responses from the
 responding replicas are not the same, we pick the replica that has the largest TS, take
 its value and send a new write command to the replicas that are out of date with the
 most updated value.

 Since we need to know the exact keys, fields and values while writing with metadata or
 while repairing, we do a "post-parsing" step that records these in order to efficiently
 populate the arguments for the Lua script that we will send to Redis on behalf of the
 read/write. (see post_parse_msg())
 This may change if we modify the parser itself to record these tokens efficiently.
 TODO: Support parsing of optional fields.

 The Lua scripts that we use as templates for supported commands are listed below as
 compile time constants.

 This is still in the beta stage and has some limitations:
 - Limited command support due to parser limitations:
     - Sorted sets and lists need optional command parsing support.
 - Repairing on the read path vs. the background has perf implications. In the future,
   all the heavy lifting will happen in the background in order to have a 0 perf
   overhead on the read and write paths.


shailesh33:
might be a good idea to port previous description into this one?

smukil:
> might be a good idea to port previous description into this one?

Good point. Done.

PR #673, opened 2019-06-17
Support all X* (streams) commands

jkimalane:
Support command DBSIZE
Support multi-bulk replies replies


jkimalane:
Discovered few issues with this patch, new version will fix those.

PR #674, opened 2019-06-17
Support all X* (streams) commands

jkimalane:
Support command DBSIZE
Support multi-bulk replies replies


jkimalane:
Skip streams tests on GitHub's CI (env. does not support redis 5.x.x)

PR #683, opened 2019-07-12
Fix crash that occurs when same-DC peer goes down when query in flight

smukil:
We incorrectly were forwarding errors when unable to connect to cross-rack
peers under DC_ONE. We should just log and ignore such errors under DC_ONE.
How this ultimately caused an issue was that the forwarded error
is serviced before any other replica can respond, which drops the request
from the client's out-q and frees it.
Then, once the replicas start responding, their server in-q's still have
a reference to the original request that got freed. Since the client has
already freed it, the other responses access garbage/repurposed memory,
hence causing a crash in a couple of places.

This patch just ignores such errors and does not forward them.


smukil:
Thanks to @shailesh33 for help with debugging.

Issue #691, opened 2019-07-30
Connection leak or connections not closing/released/recycled when client app is down


balajivenki:
Hi @shailesh33  and @ipapapa 

I am running dynomite vdynomite-v0.6.11-dirty and my dynomite conf is

`
 dyn_o_mite:
  auto_eject_hosts: true
  data_store: 0
  datacenter: DC2
  dyn_listen: 0.0.0.0:50555
  dyn_seed_provider: simple_provider
  dyn_seeds:
  - dynomit02.*****.com:50555:rack02:DC2:858993459
  - dynomit04.*****.com:50555:rack02:DC2:1717986918
  - dynomit06.*****.com:50555:rack02:DC2:2576980377
  - dynomit08.*****.com:50555:rack02:DC2:3435973836
  - dynomit10.*****.com:50555:rack02:DC2:4294967295
  - dynomit01.*****.com:50555:rack01:DC2:858993459
  - dynomit03.*****.com:50555:rack01:DC2:1717986918
  - dynomit07.*****.com:50555:rack01:DC2:3435973836
  - dynomit09.*****.com:50555:rack01:DC2:4294967295
  enable_gossip: false
  env: network
  listen: 0.0.0.0:50556
  mbuf_size: 16384
  pem_key_file: /path/dynomite.pem
  rack: rack01
  read_consistency: dc_one
  secure_server_option: datacenter
  server_failure_limit: 3
  server_retry_timeout: 30000
  servers:
  - 127.0.0.1:50180:1
  timeout: 150000
  tokens: 2576980377
  write_consistency: dc_one
`

curl -k localhost:22222/info | grep client_connections shows 60k connections made to the node and when I did netstat found a particular node(which is a docker swarm node running multiple containers of apps using dynomite) having 29k established connections. Within the app when we checked they have very few connections made to the server.

So we brought down the swarm node and moved all containers to the other worker node. Now we expect dynomite server should recycle and release those connections. However it is not, the actual swarm worker node is down and all containers are moved to other workers but dynomite still holds on to the connections made when it was alive around 29k.

This causes problem when the connection exceeds 100k and fd limit exceeds eventually crashing the nodes.

Is there any setting we are missing or is there an known issue with connection recycling with dynomite?

netstat -antp | grep :50556 | grep ESTABLISHED | wc -l shows 60k connections and among that 29k are from the swarm node we stopped long time ago. We have to restart the redis/dynomite node to go back to 0 connections.

Please help.

[...]

PR #692, opened 2019-07-31
added keepalive for dynomite listen connection

TysonHeart:
Issue #691 Enabled keepalive for listen connection of dynomite.

### DIFF
--- a/src/dyn_connection.c
+++ b/src/dyn_connection.c
@@ -234,6 +234,13 @@ rstatus_t conn_listen(struct context *ctx, struct conn *p) {
     return DN_ERROR;
   }
 
+  status = dn_set_keepalive(p->sd, 1);
+  if (status != DN_OK) {
+      log_error("set keepalive on p %d on addr '%.*s' failed: %s", p->sd,
+                p->pname.len, p->pname.data, strerror(errno));
+      // Continue since this is not catastrophic
+  }
+
   status = conn_event_add_conn(p);
   if (status < 0) {
     log_error("event add conn p %d on addr '%.*s' failed: %s", p->sd,


balajivenki:
@shailesh33  and @ipapapa  Can you please review this PR for the ticket #691

smukil:
@balajivenki Thanks for the contribution. I will have a look at it shortly.

balajivenki:
@shailesh33  and @ipapapa  Can you please review the changes? It will help us if it gets merged in the future releases.

smukil:
@balajivenki @TysonHeart The code looks good to me and is doing the right thing. One additional thing I would note is whether you want to add an option for TCP_KEEPINTVL. The default on most linux kernels for the interval between 2 keep alive probes is 7200 seconds (2 hours), which I think is a fair interval for the purpose of recycling connections for our use cases. Setting the interval too less can add a lot of load to the server, especially when you're dealing with 50k+ connections, and leaving it too large can leave a lot of un-recycled connections.

If you'd like to add a configurable option for TCP_KEEPINTVL for your use cases, feel free to submit another patch. But this can go in as is for now. Thanks for the contribution again.

http://man7.org/linux/man-pages/man7/tcp.7.html
_tcp\_keepalive\_time` (integer; default: 7200; since Linux 2.2)
The number of seconds a connection needs to be idle before TCP begins sending out keep-alive probes.  Keep-alives are sent only when the SO_KEEPALIVE socket option is enabled.  The default value is 7200 seconds (2 hours).  An idle connection is terminated after approximately an additional 11 minutes (9 probes an interval of 75 seconds apart) when keep-alive is enabled._

balajivenki:
@smukil Thanks for merging this. We thought of making it as configurable and decided to honor the values in sysctl.conf

net.ipv4.tcp_keepalive_time
net.ipv4.tcp_keepalive_intvl
net.ipv4.tcp_keepalive_probes

And we have changed our sysctl.conf to our ideal values. That way there is no confusion and all conf's related to tcp keepalive are at the same place.

murthy-chelankuri:
Can we merge this change to our latest stable version as well v0.7.0 may be as v0.7.1? We are also facing a similar issue in our prod environment

PR #693, opened 2019-08-07
Build binary tarball

msf:
Create a container for ubuntu-18.04 that serves two purposes:
- have a container version of dynomite for this OS/image
- create a binary tarball for later deployments on other instances/machines.


msf:
sorry, this was a bad pull request.

PR #694, opened 2019-08-20
Treat each rack continuum as an array

smukil:
Each rack has a 'struct continuum' pointer in it. This is used to determine
which token belongs to which node (see dyn_vnode.c).

While it has only a single pointer, the code uses it as though it's
an array of struct pointers. This potentially could lead to a lot of
confusion.

This patch addresses this by explicitly changing the member to an array
of structs.

Manually tested to make sure that there is no change to how the keys are
sharded.

### DIFF
--- a/src/dyn_core.c
+++ b/src/dyn_core.c
@@ -50,7 +50,8 @@ static void core_print_peer_status(void *arg1) {
       struct rack *rack = array_get(&dc->racks, rack_index);
       uint8_t i = 0;
       for (i = 0; i < rack->ncontinuum; i++) {
-        struct continuum *c = &rack->continuum[i];
+        struct continuum *c = (struct continuum*) array_get(&rack->continuums, i);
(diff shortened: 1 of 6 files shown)


tharanga (approved):

src/dyn_core.c
```diff
@@ -50,7 +50,7 @@ static void core_print_peer_status(void *arg1) {
       struct rack *rack = array_get(&dc->racks, rack_index);
       uint8_t i = 0;
       for (i = 0; i < rack->ncontinuum; i++) {
-        struct continuum *c = &rack->continuum[i];
+        struct continuum *c = (struct continuum*) array_get(&rack->continuums, i);
```
tharanga, line 53: Assert for null, here and all other places.

smukil: Good point. Done.

PR #710, opened 2019-09-26
Fixes msg memory leak

kjlaw89:
Frees msg->args when message is done


anujphadke (commented):
The change looks good.

anujphadke (approved):

PR #763, opened 2020-02-28
Validating florida_provider get_seeds api response 

Kevzz81517:
According to https://www.ietf.org/rfc/rfc2616.txt [6.1.1 Status Code and Reason Phrase]  The Status-Code element is a 3-digit integer result code


Kevzz81517:
Hi @ipapapa , @shailesh33  I tried create my own service for florida_provider /REST/v1/admin/get_seeds but the dynomite node couldn't get the seeds as the validation was failing. Then I changed the "200 OK" to "200" and it worked. Please review it once.

ipapapa:
@diegopacheco should probably provide some context as we are not using this part internally..

PR #775, opened 2020-08-21
Fix building and linking with GCC 10

dmolik:
dyn_ring_queue.h defines 2 global variables that are used in many
source files. This patch makes those variables ( C2G_InQ, C2G_OutQ )
external in the header, and intiates them in dyn_ring_queue.c

### DIFF
--- a/src/dyn_ring_queue.c
+++ b/src/dyn_ring_queue.c
@@ -11,6 +11,8 @@
 #include "dyn_gossip.h"
 #include "dyn_token.h"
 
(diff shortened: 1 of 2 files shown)


eskies:
This fix helps me when building the Dynomite  (Multiple definition errors both variables) on alpine Linux (ARM64 and AMD64).
Glad found the solution here, thank you @dmolik

mcouillard:
Wonderful, thank you! This also resolves the build on RHEL 9, GCC 11.

PR #809, opened 2022-07-19
Fixes crash on large payloads for MGET/MSET/SCAN

akashdeepgoel:
Fixes the issue with payloads which exceed the size of a single MBUF causing dynomite server to crash.

We now return DYNOMITE_PAYLOAD_TOO_LARGE with an exception back to the client.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,12 @@
 language: c
 script: bash ./travis.sh
 
+before_install:
+  - sudo pip3 install --upgrade pip
+  
 addons:
   apt:
+   update: true
    packages:
-   - python3
+   - python3.9
    - python3-pip
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,13 @@ language: c
 script: bash ./travis.sh
 
 before_install:
-  - sudo pip3 install --upgrade pip
+  - sudo apt remove --purge python3-pip
+  - curl -O https://bootstrap.pypa.io/pip/3.5/get-pip.py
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -1,8 +1,12 @@
 language: c
 script: bash ./travis.sh
 
+before_install:
```
anujphadke, line 4: Can you move this in a separate  commit as this is probably not related to the crash?

src/proto/dyn_redis.c
```diff
@@ -3030,7 +3030,9 @@ static rstatus_t redis_copy_bulk(struct msg *dst, struct msg *src, bool log) {
   }
 
   p = mbuf->pos;
-  ASSERT(*p == '$');
+  if (*p != '$') {
```
anujphadke, line 3033: Not related to your change, but can p be NULL at 3032? Do you think it makes sense to check for that?

.travis.yml
akashdeepgoel: This is important for the automated travis build on github as this ensures correct pip version for the build

src/proto/dyn_redis.c
akashdeepgoel: All mbufs need to come from allocation code path which ensures mbuf->pos is set to the starting of the buffer. Ideally, this should never be NULL

pkarumanchi9 (approved):
changes look good.

anujphadke (approved):

